PR #5, opened 2023-10-25
Add tag support to Oracle logs

dtru-ddog:
**Purpose**
As requested by customers we're adding support for user specified tags to logs sent to datadog intake.

**Changes**
Cleanup
- Remove unused imports
- Fix formatting
- Add .gitignore file
- Make try/catch blocks more specific
- Add more specific error handling
- Fix whitespacing, line length according to PEP8 rules
- Adding comments for context
- Adding more specific error messaging

Primary Changes
- Add support for `ddtags` option to add associated tags to logs as documented in [docs](https://docs.datadoghq.com/api/latest/logs/#send-logs)


Object Store/func.py
```diff
@@ -1,67 +1,85 @@
...
+        return
+
+    bucket = additional_details.get("bucketName")
+    if not bucket:
+        logging.getLogger().error("No bucket provided")
+        return
+
+    resource_name = data.get("resourceName")
+    if not resource_name:
+        logging.getLogger().error("No obj provided")
```
feng-andrew-ge, line 33: nit: should we use resourcename or Obj in the logger msg?

Object Store/func.py
```diff
@@ -1,67 +1,85 @@
...
+        logging.getLogger().error(err_msg)
 
-            
- 
-        headers = {'Content-type': 'application/json', 'DD-API-KEY': datadogtoken}
-        x = requests.post(datadoghost, data = json.dumps(payload), headers=headers)
-        logging.getLogger().info(x.text)
-        print(x.text)
+    for lines in data.splitlines():
+        logging.getLogger().info("lines %s", lines)
```
feng-andrew-ge, line 57: What's the default log level? Will INFO level lead to excessive log? This may print out all logs

Object Store/func.py
```diff
@@ -1,67 +1,85 @@
...
-    get_obj = object_storage_client.get_object(namespace, bucket_name, object_name)
+def request_one_object(namespace: str, bucket: str, resource_name: str) -> bytes:
+    """
+    Calls OCI to request object from Object Storage Client and decompress
+    """
+    oci_signer = oci.auth.signers.get_resource_principals_signer()
+    os_client = oci.object_storage.ObjectStorageClient(config={},
+                                                       signer=oci_signer)
+    get_obj = os_client.get_object(namespace, bucket, resource_name)
     bytes_read = gzip.decompress(get_obj.data.content)
```
feng-andrew-ge, line 84: I may not have all the context, curious: Do we handle any error/excception here caused by the client call?

Service Connector  Hub/func.py
```diff
@@ -1,49 +1,58 @@
...
+    if dd_tags:
+        payload.update({'ddtags': dd_tags})
 
+    # Invoke Datadog API with the payload.
+    # If the payload contains more than one log
+    # this will be ingested at once.
+    try:
+        headers = {'Content-type': 'application/json', 'DD-API-KEY': dd_token}
+        x = requests.post(dd_host, data=json.dumps(payload), headers=headers)
+        logging.getLogger().info(x.text)
```
feng-andrew-ge, line 44: should we use Debug level here?

Object Store/func.py
```diff
@@ -1,67 +1,85 @@
...
     try:
         body = json.loads(data.getvalue())
-        data = body.get("data", {})
-        additional_details = data.get("additionalDetails", {})
-        namespace = additional_details.get("namespace")
-        bucket = additional_details.get("bucketName")
-        obj = data.get("resourceName")
-        eventtime = body.get("eventTime")
+    except (Exception, ValueError) as ex:
+        logging.getLogger().info(str(ex))
```
feng-andrew-ge, line 15: shouldn't it be error() instead of info()?

Object Store/func.py
```diff
@@ -1,67 +1,85 @@
...
+        payload.update({"time": event_time})
+        payload.update({"service": service})
+        payload.update({"event": lines})
 
+    try:
+        headers = {'Content-type': 'application/json', 'DD-API-KEY': dd_token}
+        req = requests.post(dd_host, data=json.dumps(payload), headers=headers)
     except (Exception, ValueError) as ex:
-#        print(str(ex))
         logging.getLogger().info(str(ex))
```
feng-andrew-ge, line 70: Error() level?

Object Store/func.py
```diff
@@ -1,67 +1,85 @@
...
 import json
-import sys
-import requests
 import logging
-import time
-import gzip
 
-from fdk import response
+import oci
+import requests
```
bengillmandd, line 8: nit: Python imports should be grouped by standard lib, third party, then local ([source](https://peps.python.org/pep-0008/#imports))

Object Store/func.py
bengillmandd: Better yet, [logger.exception](https://docs.python.org/3/library/logging.html#logging.Logger.exception) will output a stack trace so there is no need to do `str(ex)`

Object Store/func.py
```diff
@@ -1,67 +1,85 @@
...
+def handler(ctx, data: io.BytesIO = None) -> None:
     try:
         body = json.loads(data.getvalue())
-        data = body.get("data", {})
-        additional_details = data.get("additionalDetails", {})
-        namespace = additional_details.get("namespace")
-        bucket = additional_details.get("bucketName")
-        obj = data.get("resourceName")
-        eventtime = body.get("eventTime")
+    except (Exception, ValueError) as ex:
```
bengillmandd, line 14: I believe [Exception](https://docs.python.org/3/library/exceptions.html#Exception) is a base class ontop of which other built in exceptions are created. This will probably catch all exceptions. If this is desired, then there is no need for ValueError here

Object Store/func.py
```diff
@@ -1,67 +1,85 @@
...
-        datadoghost = os.environ['DATADOG_HOST']
-        datadogtoken = os.environ['DATADOG_TOKEN']
+    event_time = body.get("eventTime")
 
-        for lines in data.splitlines():
-            logging.getLogger().info("lines " + lines)
-            payload = {}
-            payload.update({"host":obj})
-            payload.update({"time": eventtime})
+    source = "Oracle Cloud"  # Adding a source name.
```
bengillmandd, line 38: nit: constants in python are written as all caps snake case 
```suggestion
    SOURCE = "Oracle Cloud"  # Adding a source name.
```

Object Store/func.py
```diff
@@ -1,67 +1,85 @@
...
+        payload.update({"ddtags": dd_tags})
+        payload.update({"host": resource_name})
+        payload.update({"time": event_time})
+        payload.update({"service": service})
+        payload.update({"event": lines})
 
+    try:
+        headers = {'Content-type': 'application/json', 'DD-API-KEY': dd_token}
+        req = requests.post(dd_host, data=json.dumps(payload), headers=headers)
     except (Exception, ValueError) as ex:
```
bengillmandd, line 69: nit: similar to above, Exception here will catch everything and ValueError is not needed

Object Store/func.py
```diff
@@ -1,67 +1,85 @@
...
+        return
+
+    resource_name = data.get("resourceName")
+    if not resource_name:
+        logging.getLogger().error("No obj provided")
+        return
 
-        datadoghost = os.environ['DATADOG_HOST']
-        datadogtoken = os.environ['DATADOG_TOKEN']
+    event_time = body.get("eventTime")
```
bengillmandd, line 36: Are we ok to have a potentially Null event_time?

Object Store/func.yaml
```diff
@@ -6,5 +6,6 @@ entrypoint: /python/bin/fdk /function/func.py handler
 memory: 1024
 timeout: 120
 config:
-  DATADOG_HOST: https://http-intake.logs.datadoghq.com/v1/input
-  DATADOG_TOKEN: <DATADOG_API_TOKEN>
+  DATADOG_HOST: https://http-intake.logs.datadoghq.com/v1/input     # DD Log Intake Host
+  DATADOG_TOKEN: <DATADOG_API_TOKEN>                                # DD API Token
+  # DATADOG_TAGS: "prod:true"                                        # Tags associated with logs
```
bengillmandd, line 11: I dont have context here but should this be commented out?

Service Connector  Hub/func.py
```diff
@@ -1,49 +1,58 @@
...
 
-def process(body):
+import requests
+
+
+def process(body: dict) -> None:
+    data = body.get("data", {})
+    source = body.get("source")
+    time = body.get("time")
+    dd_source = "oracle_cloud"
```
bengillmandd, line 13: nit: consts should be all caps

Service Connector  Hub/func.py
```diff
@@ -1,49 +1,58 @@
...
 
-"""
-This function receives the logging json and invokes the Datadog endpoint for ingesting logs. https://docs.cloud.oracle.com/en-us/iaas/Content/Logging/Reference/top_level_logging_format. htm#top_level_logging_format
-If this Function is invoked with more than one log the function go over each log and invokes the Datadog endpoint for ingesting one by one.
-"""
-def handler(ctx, data: io.BytesIO=None):
+def handler(ctx, data: io.BytesIO = None) -> None:
+    """
+    This function receives the logging json and invokes the Datadog endpoint
+    for ingesting logs. https://docs.cloud.oracle.com/en-us/iaas/Content/Logging/Reference/top_level_logging_format.htm#top_level_logging_format
```
bengillmandd, line 52: thank you for this 😅

Service Connector  Hub/func.py
```diff
@@ -1,49 +1,58 @@
...
+        logging.getLogger().error(err_msg)
+    if dd_tags:
+        payload.update({'ddtags': dd_tags})
 
+    # Invoke Datadog API with the payload.
+    # If the payload contains more than one log
+    # this will be ingested at once.
+    try:
+        headers = {'Content-type': 'application/json', 'DD-API-KEY': dd_token}
+        x = requests.post(dd_host, data=json.dumps(payload), headers=headers)
```
bengillmandd, line 43: nit: this should be named something like `response` rather than `x`

Service Connector  Hub/func.yaml
```diff
@@ -7,4 +7,5 @@ memory: 1024
 timeout: 120
 config:
   DATADOG_HOST: https://http-intake.logs.datadoghq.com/v1/input 
-  DATADOG_TOKEN: <DATADOG_API_TOKEN>
\ No newline at end of file
+  DATADOG_TOKEN: <DATADOG_API_TOKEN>
+  # DATADOG_TAGS: "prod:true"                                        # Tags associated with logs
```
bengillmandd, line 11: should this be commented out?

bengillmandd (commented):
overall looks good, I just had mostly nit comments. Thanks for taking care of this for us

Service Connector  Hub/func.py
kanishktripathi: Or it can be initialized at module like: `_DD_SOURCE = "oracle_cloud"`

Object Store/func.py
```diff
@@ -1,67 +1,85 @@
...
+        payload.update({"ddsource": source})
+        payload.update({"ddtags": dd_tags})
+        payload.update({"host": resource_name})
+        payload.update({"time": event_time})
+        payload.update({"service": service})
+        payload.update({"event": lines})
 
+    try:
+        headers = {'Content-type': 'application/json', 'DD-API-KEY': dd_token}
+        req = requests.post(dd_host, data=json.dumps(payload), headers=headers)
```
kanishktripathi, line 68: nit: should `req` be initialized if not using the variable. If using this, should `res`/`response` be a better option?

Object Store/func.py
dtru-ddog: Agreed it's not functional, was more marking that it was known that the `ValueError` would pop up. But can update accordingly minor detail.

Object Store/func.py
dtru-ddog: Overall agreed, was attempting to not change how logs were printed in this implementation. But can take a stab at updating and also add the logger at the module level.

Object Store/func.py
dtru-ddog: log msg already existed, I can update to better match params.

Object Store/func.py
dtru-ddog: Default logging level is WARNING per [docs](https://docs.python.org/3/howto/logging.html)

Object Store/func.py
dtru-ddog: Not at the moment, this was meant to be a cleanup of the current execution without adding too much additional logic to change behavior. There's probably some error catching worth doing here, but will extend that for future work.

Object Store/func.yaml
dtru-ddog: This was commented out since this is meant to be user supplied tags that they can specify. I commented out here to provide an example of tags the user can specify in their logs.

Service Connector  Hub/func.py
dtru-ddog: we default to warn so info should be fine here.

Service Connector  Hub/func.yaml
dtru-ddog: Similar to the other one it's just a customer option. Meant to be commented out for customer to specify their own custom tags.

Object Store/func.py
dtru-ddog: should be grouped like this now with oci + requests being the 3rd party imports

Object Store/func.py
dtru-ddog: that's a valid point will add another check here

kanishktripathi (approved):

PR #7, opened 2024-08-23
ECI-221 Add default namespaces when creating connector hub

kanishktripathi:
* Add default namespaces when creating connector hub
* Currently limited to 50 total namespaces for a single connector hub


natwilkinson (approved):
ty!

PR #10, opened 2024-09-27
ECI-161 Fix ddsource in log forwarding function

kanishktripathi:
**what:**
* Change source constant to `oracle_cloud` in log forwarding function
* In service connector log forwarding function, change `ddsource` field value to `oracle_cloud`

**why:**
* The log forwarding takes the the `ddsource` value as `oracle_cloud` without that the logs are not accepted

### DIFF
--- a/Service Connector  Hub/func.py
+++ b/Service Connector  Hub/func.py
@@ -7,7 +7,7 @@ import requests
 
 logger = logging.getLogger(__name__)
 
-DD_SOURCE = "Oracle Cloud"  # Adding a source name.
+DD_SOURCE = "oracle_cloud"  # Adding a source name.
 DD_SERVICE = "OCI Logs"  # Adding a service name.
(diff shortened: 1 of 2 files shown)


Service Connector  Hub/func.py
```diff
@@ -22,7 +22,7 @@ def process(body: dict) -> None:
     payload.update({"source": source})
     payload.update({"time": time})
     payload.update({"data": data})
-    payload.update({"ddsource": DD_SERVICE})
+    payload.update({"ddsource": DD_SOURCE})
```
sva91, line 25: This might require update in the test function as well, right ?

kanishktripathi: The test is already updated with this value.

sva91: Sounds good.

PR #15, opened 2024-10-21
Service Connector deployed to chosen compartment with compartment specific metric namespaces

ecout:
As per Support Case: https://help.datadoghq.com/hc/en-us/requests/1893022?page=1 there max limits in the Service Connector Compartment: a maximum of 5 Compartments and a maximum metric namespace of 50. 
In this pull request instead of deploying the service connector resource to monitor the root/tenancy compartment, switch it to monitor compartment where the resources live using the data source oci_monitoring_metrics grouped by namespace to extract the supported namespaces from the chosen compartment by compartment id rather than the local.connector_metric_namespaces hardcoded input.


kanishktripathi:
This change was incorporated in this release
https://github.com/DataDog/oracle-cloud-integration/releases/tag/datadog-oci-orm-v1.3.0

PR #21, opened 2024-12-05
ECI-394 Add docker login and function app modules

sva91:
## What
- Added Function App module
- Added container registry module, and implemented it partially
1. Within container registry module, added functionality to perform docker login. There is still more functionality that will be added in separate PR, to create docker repository, build function image, and push image.

Enhancements:
1. No longer require users to specify username and password for docker. Username will be fetched automatically from ORM, using user ocid.
2. A user can choose to auto-generate an auth token, or provide a pre-existing token.
3. Added resiliency to check if a user already has 2 auth tokens. OCI has a restriction of 2 on number of auth tokens that can be generated by a user. In addition, attempting docker login multiple times to reduce probability of failure on the login step.
4. Containerregistry module is only called if a user chooses to build a new function image.

## Why

1. OCI Log forwarding infrastructure
2. Make docker login experience seamless

## Testing
- Tested using ORM : https://cloud.oracle.com/resourcemanager/jobs/ocid1.ormjob.oc1.iad.amaaaaaaehwejaiayz5giefi4lasrixftpgf6hbk5dnmmshe5cu2fjprz4jq?region=us-ashburn-1
- Tested by running terraform scripts directly


datadog-logs-oci-orm/modules/containerregistry/login.tf
```diff
@@ -0,0 +1,91 @@
...
+                # Step 2: Create a new auth token
+                echo "Creating a new auth token..."
+                new_token_value=$(oci iam auth-token create --user-id ${var.current_user_ocid} \
+                    --description "${var.auth_token_description}" --query "data.token" --raw-output)
+
+                # Step 3: Sleep for 30 seconds to ensure token propagation
+                echo "Waiting for 30 seconds to ensure token availability..."
+                sleep 30
+
+                echo $new_token_value > /tmp/new_auth_token.txt
```
bengillmandd, line 56: nit ⛏️  I'm curious on what infra does this text file gets saved? Reason I bring it up is because if for some reason, the next resource call doesn't run and it doesn't get deleted, it could be a security risk

datadog-logs-oci-orm/modules/functionapp/variables.tf
```diff
@@ -0,0 +1,56 @@
...
+  default     = {}
+}
+
+variable "resource_name_prefix" {
+  type        = string
+  description = "The prefix for the name of all of the resources"
+}
+
+#*************************************
+#         Function Application Variables
```
bengillmandd, line 18: Massive nit pick but this is off center
```suggestion
#     Function Application Variables
```

datadog-logs-oci-orm/variables.tf
```diff
@@ -39,3 +39,66 @@ variable "tenancy_ocid" {
   type        = string
   description = "OCI tenant OCID, more details can be found at https://docs.cloud.oracle.com/en-us/iaas/Content/API/Concepts/apisigningkey.htm#five"
 }
+variable "current_user_ocid" {
+  type        = string
+  description = "OCID of the logged in user running the terraform script"
+}
+
+#*************************************
+#         Function Application Variables
```
bengillmandd, line 48:
```suggestion
#     Function Application Variables
```

datadog-logs-oci-orm/variables.tf
```diff
@@ -39,3 +39,66 @@ variable "tenancy_ocid" {
...
+}
+
+variable "datadog_tags" {
+  type        = string
+  default     = ""
+  description = "The tags to be sent with the logs. The tags should be in the format key1:value1,key2:value2"
+}
+
+#*************************************
+#         Container Registry Variables
```
bengillmandd, line 90:
```suggestion
#     Container Registry Variables
```

datadog-logs-oci-orm/main.tf
```diff
@@ -13,3 +13,25 @@ module "policy" {
...
+}
+
+module "containerregistry" {
+    source = "./modules/containerregistry"
+    oci_region_key = local.oci_region_key
+    tenancy_ocid = var.tenancy_ocid
+    current_user_ocid = var.current_user_ocid
+    auth_token_description = var.auth_token_description
+    auth_token = var.auth_token
+    count = var.function_image_path == "" ? 1 : 0
```
bengillmandd, line 36: Maybe I am missing something but I don't see this variable being used. And maybe there is a better name for it than "count"?

bengillmandd (commented):
Overall lgtm. I'm not super familiar with terraform so I don't have too much to add and sorry for some potentially dumb questions. 

This PR is missing a link to Jira ticket (though I can guess which ticket it belongs to)

Maybe I was missing something but I couldn't see where in the code this functionality was implemented: _"Containerregistry module is only called if a user chooses to build a new function image."_

datadog-logs-oci-orm/modules/containerregistry/login.tf
```diff
@@ -0,0 +1,91 @@
+resource "null_resource" "get_username" {
```
kanishktripathi, line 1: we can't use the terraform [resource for this](https://registry.terraform.io/providers/oracle/oci/latest/docs/data-sources/identity_user) ?

sva91: I think we should be able to. I will experiment with that, and change it if that works.

datadog-logs-oci-orm/main.tf
sva91: count is basically a meta-argument defined by the terraform : https://developer.hashicorp.com/terraform/language/meta-arguments/count
It represents number of infrastructure objects we expect to create from a resource definition.

datadog-logs-oci-orm/modules/containerregistry/login.tf
sva91: updated

datadog-logs-oci-orm/main.tf
bengillmandd: ouch

bengillmandd (approved):

PR #23, opened 2024-12-26
Add detailed logging option to metrics function

rheei:
**what:**
Add option for OCI customers to view their raw payloads from OCI
**why:**
Support debugging
**testing:**

### DIFF
--- a/datadog-oci-orm/metrics-setup/metrics-function/func.py
+++ b/datadog-oci-orm/metrics-setup/metrics-function/func.py
@@ -14,6 +14,7 @@ logger = logging.getLogger(__name__)
 
 
 OUTPUT_MESSAGE_VERSION = "v1.0"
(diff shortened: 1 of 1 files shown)


datadog-oci-orm/metrics-setup/metrics-function/func.py
```diff
@@ -15,6 +15,9 @@
 
 OUTPUT_MESSAGE_VERSION = "v1.0"
 
+# Switch value to True to get payload logs 
+DETAILED_LOGGING_ENABLED = False
```
natwilkinson, line 19: Can we add this as an environment var, similar to how we do for `DD_INTAKE_HOST` and `DD_API_KEY`? That way turning this off/on doesn't require a new function image/deployment.

natwilkinson (approved):

PR #24, opened 2024-12-26
ECI-395 Search Resources for Logging

sva91:
## What

- Add Exclude_Services multi-select enum for users to skip forwarding logs from some services
- Add Compartments_CSV file upload option to enable users specify compartments for log forwarding. By using csv, we have enabled compartment level configurations by the user.
- For all supported logging services that aren't excluded, identify resources in all compartments specified by the user.

## Why

- Enable log forwarding in OCI

## Testing

- Only Logging Module Test: https://cloud.oracle.com/resourcemanager/stacks/ocid1.ormstack.oc1.iad.amaaaaaaehwejaiaasomsmas7k7t67feoqexk4xzatudipbk2hwrtdgfnoia?region=us-ashburn-1
- End to end test so far : https://cloud.oracle.com/resourcemanager/stacks/ocid1.ormstack.oc1.iad.amaaaaaaehwejaiapg44gaenwqjnoejyrlsggbljpktqwp4tzi7yccud5qiq?region=us-ashburn-1

## Docs Notes

- It might be preferred to separate the plan and apply stage. Plan stage will identify all resources from which logs should be forwarded.
- When applying a stack, select a previously successful plan job, instead of applying from beginning so that we can separate resource discovery from resource creation.


datadog-logs-oci-orm/modules/resourcediscovery/variables.tf
```diff
@@ -0,0 +1,14 @@
+variable "group_id" {
```
bengillmandd, line 1: I'm not sure if "service_id" or "service_name"? might be a better name for this? I see "group_id" as an OCI Group's OCID (eg. `DatadogAuthGroup`)

bengillmandd (approved):
Overall lgtm. Had a few nit picks. Would be nice to see what the customer UI experience would be for this if it's possible

datadog-logs-oci-orm/modules/resourcediscovery/variables.tf
sva91: The purpose of this variable is to specify `Unique Name for the group of resource types`. It doesn't really have to be associated to a service id. The purpose here is to provide any string (it may not even be a service id).

This ensures that the abstraction for any caller code to use `resourcediscovery` module is as simple as possible (all you need to provide is compartment ocid and list of resource types).

PR #27, opened 2025-01-06
ECI [OCI] allow datad0g to be metric intake endpoint

rheei:
**what:**
Allow staging to have it's own metric intake endpoint but don't show it in the drop down menu
**why:**
So we can create metric stack from staging
**testing:**

### DIFF
--- a/datadog-oci-orm/metrics-setup/variables.tf
+++ b/datadog-oci-orm/metrics-setup/variables.tf
@@ -36,7 +36,7 @@ variable "datadog_environment" {
   description = "The endpoint to hit for sending the metrics. Varies by different datacenter"
   validation {
     condition = contains(["ocimetrics-intake.datadoghq.com", "ocimetrics-intake.us5.datadoghq.com", "ocimetrics-intake.us3.datadoghq.com",
-    "ocimetrics-intake.datadoghq.eu", "ocimetrics-intake.ap1.datadoghq.com", "ocimetrics-intake.ddog-gov.com"], var.datadog_environment)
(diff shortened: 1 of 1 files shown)


natwilkinson (approved):

datadog-oci-orm/metrics-setup/variables.tf
```diff
@@ -36,7 +36,7 @@ variable "datadog_environment" {
   description = "The endpoint to hit for sending the metrics. Varies by different datacenter"
   validation {
     condition = contains(["ocimetrics-intake.datadoghq.com", "ocimetrics-intake.us5.datadoghq.com", "ocimetrics-intake.us3.datadoghq.com",
-    "ocimetrics-intake.datadoghq.eu", "ocimetrics-intake.ap1.datadoghq.com", "ocimetrics-intake.ddog-gov.com"], var.datadog_environment)
+    "ocimetrics-intake.datadoghq.eu", "ocimetrics-intake.ap1.datadoghq.com", "ocimetrics-intake.ddog-gov.com", "ocimetrics-intake.datad0g.com"], var.datadog_environment)
```
sva91, line 39: Don't we need a change in schemas.yaml file ?

rheei: @sva91 we don't want customers to be able to choose this option for their endpoint because it is internal. When t[his web-ui PR](https://github.com/DataDog/web-ui/pull/176211/files) goes out, in theory it will be able to be populated with the staging site even though it is not in the drop down.

sva91: I agree customers shouldn't be able to see the option. Have you tested this ? This field is defined as an [enum](https://github.com/DataDog/oracle-cloud-integration/blob/51a5ada6f25880c1b92e219ee1149fe3d73e78f9/datadog-oci-orm/metrics-setup/schema.yaml#L96). Is it able to accept an unsupported value ?

PR #29, opened 2025-01-13
ECI [OCI] include more log attributes and assign "DD_SOURCE" based on OCI service

rheei:
We need to forward the "oracle" and "type" attributes from OCI logs to get all the information for facets.
Also, we need to be able to distinguish what type of log we are sending from OCI to dd, so this PR checks if the log is a service log. If it is not, it extracts the service name from the "type" attribute.

### DIFF
--- a/datadog-logs-oci-orm/logs-function/func.py
+++ b/datadog-logs-oci-orm/logs-function/func.py
@@ -7,7 +7,7 @@ import requests
 
 logger = logging.getLogger(__name__)
 
-DD_SOURCE = "OCI Logs"  # Adding a source default name. The source will be mapped to the log's respective service to be processed by the correct pipeline in Datadog. 
+DD_SOURCE = "oci.logs"  # Adding a source default name. The source will be mapped to the log's respective service to be processed by the correct pipeline in Datadog. 
 DD_SERVICE = "oci"  # Adding a service name.
 DD_TIMEOUT = 10 * 60  # Adding a timeout for the Datadog API call.
 DD_BATCH_SIZE = 1000  # Adding a batch size for the Datadog API call.
--- a/datadog-logs-oci-orm/logs-function/func.py
+++ b/datadog-logs-oci-orm/logs-function/func.py
@@ -7,7 +7,7 @@ import requests
 
 logger = logging.getLogger(__name__)
 
-DD_SOURCE = "oci.logs"  # Adding a source default name. The source will be mapped to the log's respective service to be processed by the correct pipeline in Datadog. 
+DD_SOURCE = "oci.logs"  # Adding a source default name. The source will be mapped to the log's respective service to be processed by the correct pipeline in Datadog.
 DD_SERVICE = "oci"  # Adding a service name.
 DD_TIMEOUT = 10 * 60  # Adding a timeout for the Datadog API call.
 DD_BATCH_SIZE = 1000  # Adding a batch size for the Datadog API call.
@@ -31,10 +31,10 @@ def _process(body: list[dict]) -> None:
     This function retrieves the Datadog endpoint URL and token from environment variables,
     processes each log entry in the provided list, compresses the payload, and sends it
     to the Datadog API.
-    
+
     Args:
         body (list[dict]): A list of log entries, where each log entry is represented as a dictionary.
-    
+
     Raises:
         KeyError: If the required environment variables 'DATADOG_HOST' or 'DATADOG_TOKEN' are not set.
         Exception: If there is an error during the API request or payload processing.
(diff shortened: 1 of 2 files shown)


bengillmandd (approved):

datadog-logs-oci-orm/logs-function/func.py
```diff
@@ -76,13 +76,17 @@ def _process_single_log(body: dict) -> dict:
     data = body.get("data", {})
     source = body.get("source")
     time = body.get("time")
+    logtype = body.get("type")
+    oracle = body.get("oracle")
```
sva91, line 80: Are these 2 fields always present in logs body? If not, this might error out, so we could provide a default.

rheei: I'm pretty sure they are always there according to [this doc](https://docs.oracle.com/en-us/iaas/Content/Logging/Reference/top_level_logging_format.htm), but we can add a check if you think it would be worthwhile. @sva91

sva91: Do we get these for Audit logs as well ?

rheei: these are for all OCI logs, yes

datadog-logs-oci-orm/logs-function/func.py
```diff
@@ -71,18 +71,45 @@ def _process(body: list[dict]) -> None:
     except Exception as ex:
         logger.exception(ex)
 
+def _get_oci_source_name(body: dict) -> str:
```
sva91, line 74: Can you add tests for this please [here](https://github.com/DataDog/oracle-cloud-integration/blob/sva91/log_forwarding_automation_modules/datadog-logs-oci-orm/logs-function/tests/test_func.py) ?

datadog-logs-oci-orm/logs-function/func.py
```diff
@@ -7,8 +7,8 @@
 
 logger = logging.getLogger(__name__)
 
-DD_SOURCE = "oracle_cloud"  # Adding a source name.
-DD_SERVICE = "OCI Logs"  # Adding a service name.
+DD_SOURCE = "OCI Logs"  # Adding a source default name. The source will be mapped to the log's respective service to be processed by the correct pipeline in Datadog.
```
sva91, line 10: I think it might be better to provide an underscored string, and avoid space, since we can use this field to filter logs in Datadog.

datadog-logs-oci-orm/logs-function/func.py
```diff
@@ -71,18 +71,45 @@ def _process(body: list[dict]) -> None:
...
+    logtype = body.get("type")
+    oracle = body.get("oracle")
+    ddsource = _get_oci_source_name(body)
 
     payload = {
         "source": source,
         "timestamp": time,
         "data": data,
-        "ddsource": DD_SOURCE,
+        "ddsource": ddsource,
```
kanishktripathi, line 109: Any reason we're changing the default ddsource and service values? Are we getting logs with the changed values?

datadog-logs-oci-orm/logs-function/func.py
```diff
@@ -71,18 +71,45 @@ def _process(body: list[dict]) -> None:
...
+    """
+    oracle = body.get("oracle")
+    logtype = body.get("type")
+
+    if oracle.get("loggroupid") != None and oracle.get("loggroupid") == "_Audit":
+        return "oci.audit"
+
+    if logtype != "":
+        # logtype is of format com.oraclecloud.{service}.{resource-type}.{category}
+        return "oci." + logtype.split(".")[2]
```
kanishktripathi, line 93: Are we sure about the consistency of this pattern? What if we get index out of bound error on this. i.e. `logtype.split(".")` is not exactly like this.
IMO, we should not just assume having a correct format all the time and handle this scenario since this code will run on a customer environment.

datadog-logs-oci-orm/logs-function/func.py
rheei: Roman and I spoke about this today and the reason is that we want each type of oci log to be its own "source", that way it is routed to the correct log processing pipeline. Other clouds do this as well, where their sources are {cloud provider}.{service} to keep the processing more granular, rather than lumping all of them into one messy pipeline. @kanishktripathi

datadog-logs-oci-orm/logs-function/func.py
rheei: according to [this documentation](https://docs.oracle.com/en-us/iaas/Content/Logging/Reference/top_level_logging_format.htm), this pattern should be consistent. If there is not a type, we will default to the "oci.logs" value instead. We need to be able to determine the service before we forward the log so that it is processed by the intended pipeline.

datadog-logs-oci-orm/logs-function/func.py
rheei: Also, I have personally deployed this function in OCI and we are seeing logs with their respective sources: 
[audit](https://app.datadoghq.com/logs?query=source%3A%22oci.audit%22&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=driveline&stream_sort=time%2Cdesc&viz=stream&from_ts=1736799197400&to_ts=1736800097400&live=true), [vcn](https://app.datadoghq.com/logs?query=source%3A%22oci.vcn%22&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1736782149347&to_ts=1736783049347&live=true),[ oke](https://app.datadoghq.com/logs?query=source%3A%22oci.kubernetes%22&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1736782149347&to_ts=1736783049347&live=true
)

datadog-logs-oci-orm/logs-function/func.py
rheei: We can handle the out of bounds scenario, agreed

datadog-logs-oci-orm/logs-function/func.py
sva91: It will probably be less error-prone if we add that check.

sva91 (approved):

PR #33, opened 2025-01-28
ECI-453 [OCI] add redaction option for sensitive log fields

rheei:
**what:**
we want the ability to hide certain fields in the audit logs
**why:**
https://datadoghq.atlassian.net/browse/ECI-453
**testing:**
unit tests

### DIFF
--- a/datadog-logs-oci-orm/logs-function/func.py
+++ b/datadog-logs-oci-orm/logs-function/func.py
@@ -11,7 +11,13 @@ DD_SOURCE = "oci.logs"  # Adding a source default name. The source will be mappe
 DD_SERVICE = "oci"  # Adding a service name.
 DD_TIMEOUT = 10 * 60  # Adding a timeout for the Datadog API call.
 DD_BATCH_SIZE = 1000  # Adding a batch size for the Datadog API call.
-
+REDACTED_FIELDS = [
+    "data.identity.credentials",
+    "data.request.headers.Authorization",
+    "data.request.headers.authorization",
+    "data.request.headers.X-OCI-LB-PrivateAccessMetadata",
+    "data.request.headers.opc-principal"
+]
 
 def _should_compress_payload() -> bool:
     return os.environ.get("DD_COMPRESS", "true").lower() == "true"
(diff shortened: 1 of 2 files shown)


maxpyf (approved):

datadog-logs-oci-orm/logs-function/func.py
```diff
@@ -11,7 +11,16 @@
...
 DD_BATCH_SIZE = 1000  # Adding a batch size for the Datadog API call.
+REDACTED_FIELDS = [
+    "data.identity.credentials",
+    "data.request.headers.Authorization",
+    "data.request.headers.authorization",
+    "data.request.headers.X-OCI-LB-PrivateAccessMetadata",
+    "data.request.headers.opc-principal"
+]
 
+def _should_redact_sensitive_data() -> bool:
```
sva91, line 22: Why are we providing this option ? I think this should just be always true for everyone.

datadog-logs-oci-orm/logs-function/func.yaml
```diff
@@ -10,3 +10,4 @@ config:
   DATADOG_TOKEN: <DATADOG_API_TOKEN>                                # DD API Token
   DD_COMPRESS: "true"                                               # Enable Compression
   # DATADOG_TAGS: "prod:true"                                       # Tags associated with logs
+  REDACT_SENSITIVE_DATA: "true"                                     # Redact sensitive data
```
sva91, line 13: Similar to above comment, I don't think we need to provide this as an option customers get to choose

datadog-logs-oci-orm/logs-function/tests/test_func.py
```diff
@@ -33,7 +33,10 @@ class TestLogForwarderFunction(TestCase):
             "data": {
               "level": "INFO",
               "message": "Run succeeded - Read 0 messages from source and wrote 0 messages to target",
-              "messageType": "CONNECTOR_RUN_COMPLETED"
+              "messageType": "CONNECTOR_RUN_COMPLETED",
```
sva91, line 36: Like we discussed, please add a test case where the fields aren't present in logs.

datadog-logs-oci-orm/logs-function/func.py
rheei: Instead of letting customers choose what fields are redacted, we are still allowing the option to opt in/out of this feature. cc: @natwilkinson  might have more insight into this

PR #34, opened 2025-01-29
ECI-418 Change the auth user email for DatadogAuthUser

kanishktripathi:
**what:**
* Change the email of DatadogAuthUser to `datadoghq` domain.

**why:**
* For security reason. To avoid sending the user creation email to a known email domain
* Currently releasing with this email as a quick fix. TBC with internal compliance on the best course of action.


rheei (commented):
shouldn't it be like support@datadoghq.com? 
edit: So we just want the email to be a fake email so that the info doesn't get sent somewhere? Would we need to have the user set their own email in the future?

kanishktripathi:
> shouldn't it be like [support@datadoghq.com](mailto:support@datadoghq.com)?

That is a legit group with active users.

bengillmandd (approved):

PR #36, opened 2025-02-03
Sva91/ECI-399 E2E logs private beta update stack

sva91:
## What

- Replace exclude_services with include_services
- Replace CSV Upload for compartments with Comma separated values for compartments to align with the current metrics experience
- Update Docker inputs to align with the current metrics experience
- Remove Function Shape variable to align with current metrics experience

## Why

- Prepare for Private Beta Launch for Logs

## Testing
- Stack : [Link](https://cloud.oracle.com/resourcemanager/stacks/ocid1.ormstack.oc1.iad.amaaaaaaehwejaiauihole7chllbzqygt3t2b7tt2heh7drsvimkeulzc7bq?region=us-ashburn-1)

### DIFF
--- a/datadog-logs-oci-orm/schema.yaml
+++ b/datadog-logs-oci-orm/schema.yaml
@@ -11,7 +11,6 @@ variableGroups:
       - ${tenancy_ocid}
       - ${region}
       - ${compartment_ocid}
-      - ${current_user_ocid}
   - title: "Network options"
     variables:
       - ${create_vcn}
@@ -25,27 +24,15 @@ variableGroups:
       - ${datadog_tags}
   - title: "Function settings"
     variables:
-      - ${function_app_shape}
-      - ${create_new_function_image}
-      - ${function_image_path}
-  - title: "Container Registry"
-    variables:
-      - ${use_different_service_user}
-      - ${service_user_ocid}
-      - ${create_auth_token}
-      - ${auth_token_description}
-      - ${auth_token}
-    visible: ${create_new_function_image}
+      - ${oci_docker_username}
+      - ${oci_docker_password}
   - title: "Logging"
     variables:
-      - ${exclude_services}
+      - ${include_services}
       - ${logging_compartments_csv}
       - ${enable_audit_log_forwarding}
 
 variables:
-  current_user_ocid:
-    visible: false
-  
   tenancy_ocid:
       visible: false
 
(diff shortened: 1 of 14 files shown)


datadog-logs-oci-orm/schema.yaml
```diff
@@ -120,82 +107,29 @@ variables:
...
-        - ${create_auth_token}
+    sensitive: true
 
 #Logging
-  exclude_services:
-    title: Exclude Services from Logging
+  include_services:
+    title: Include Services for Logging
     type: enum
     description: List of services to exclude from logging.
```
bengillmandd, line 127: ⛏️  description should be updated

datadog-logs-oci-orm/schema.yaml
```diff
@@ -229,12 +162,14 @@ variables:
...
-  logging_compartments_csv:
-    title: Compartments CSV
-    description: "Upload a CSV file containing the list of compartments. The file must include a column named 'compartment_id'."
-    type: file
+  
+  logging_compartments:
+    title: Logging compartments
+    type: text
     required: true
+    description: The list of logging compartments. Enter comma separated list of OCID of compartments to monitor.
```
bengillmandd, line 170: ⛏️ These would be the list of compartments for which customers would like to send logs to DD correct? Not sure IMO that "compartments to monitor" makes that clear

datadog-logs-oci-orm/schema.yaml
```diff
@@ -207,7 +141,6 @@ variables:
       - "cloud_guard_raw_logs_prod"
       - "och"
       - "oke-k8s-cp-prod"
-      - "contentdeliverynetwork"
```
bengillmandd, line 210: curious why cdn is removed here?

datadog-logs-oci-orm/schema.yaml
```diff
@@ -120,82 +107,29 @@ variables:
...
-  
-  auth_token:
-    title: Auth Token
-    type: password
-    description: The auth token to be used for the container registry.
     sensitive: true
+  oci_docker_password:
+    title: OCI Docker registry auth token
+    type: password
+    description: The user auth token for the OCI docker container registry. Used in creating function image.
```
bengillmandd, line 119: Is this also not required if using an existing function?

datadog-logs-oci-orm/variables.tf
```diff
@@ -90,39 +66,32 @@ variable "datadog_tags" {
 }
 
 #************************************
-#    Container Registry Variables    
+#    Function setting Variables
```
bengillmandd, line 69: Function Setup Variables? This is what a comment says in the schema.yaml

bengillmandd (approved):
Just a bunch of nit ⛏️ s Given I'm not super familiar with terraform, might be good to get kanishk's review here too

datadog-logs-oci-orm/schema.yaml
sva91: Yea, that's right. I will update this.

datadog-logs-oci-orm/schema.yaml
sva91: We have taken away the option to use an existing function image. Since that's the current metrics experience, I have updated the logs to be same.

bengillmandd: Ah ok I would remove " Not required if using an existing image path." from the oci_docker_username description then

sva91: good catch. thanks.

datadog-logs-oci-orm/schema.yaml
sva91: It's because, as part of testing, CDN resources are not supported by SearchResources API, which is what the terraform also uses for resource discovery. Since, we can't discover these resources, we can't really automate logging enablement for this service. That's why I took it out as an option.

PR #38, opened 2025-02-05
ECI [OCI] add back option to change image shape for metrics

rheei:
**what:**
Allow customers to specify image shape in metric stack
**why:**
Customers in Japan regions cannot use GENERIC_ARM shape
**testing:**
Ran stack successfully

### DIFF
--- a/datadog-oci-orm/metrics-setup/schema.yaml
+++ b/datadog-oci-orm/metrics-setup/schema.yaml
@@ -22,6 +22,7 @@ variableGroups:
       - ${metrics_compartments}
   - title: "Function settings"
     variables:
+      - ${function_app_shape}
       - ${oci_docker_username}
       - ${oci_docker_password}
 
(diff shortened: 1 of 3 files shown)


datadog-oci-orm/metrics-setup/schema.yaml
```diff
@@ -172,9 +173,19 @@ variables:
...
       - "oci_waf"
       - "oracle_oci_database"
-    
+
 
   # Function setup
+  function_app_shape:
+    title: Function Application shape
+    type: enum
+    description: The shape of the function application. The docker image should be built accordingly. Use GENERIC_ARM if using Oracle Resource manager stack.
```
kanishktripathi, line 182: `Use GENERIC_ARM if using Oracle Resource manager stack`. This can be removed if not applicable.
Perhaps we can replace it with. `Select GENERIC_X86 if there's an error in creating with GENERIC_ARM.` We can also eliminate `GENERIC_X86_ARM`

rheei: agreed

kanishktripathi (approved):

PR #39, opened 2025-02-05
ECI-455 Add service account user for metrics setup and docker images

kanishktripathi:
**what:**

** Policy changes:**
* Create a new user `DatadogAuthWriteUser` which is  used in pushing docker images.
* Change the name of the groups, dynamic group, policy and users so that it doesn't collide with existing stack.
* Remove the option for the name of group name, policy and dynamic group name

**Metric stack:**
* The docker images are only created with the user login created through policy stack.
* If the user is not created through policy stack then no resource will be created.
* A new auth token is created every time on stack run so it is not required to store the credential. It takes at least 60-120 sec for the token to be accepted and used for docker login.

**why:**
* To reduce additional complexity in docker credentials.

**testing:**
* Tested on `datadog84` tenancy. See [here](https://cloud.oracle.com/resourcemanager/jobs/ocid1.ormjob.oc1.eu-frankfurt-1.amaaaaaaehwejaiaavrk7eex4gqpq6pbpvr3l3iqodihqeol7rf7pj62ufza?region=eu-frankfurt-1)

### DIFF
--- a/datadog-oci-orm/metrics-setup/auth_token.py
+++ b/datadog-oci-orm/metrics-setup/auth_token.py
@@ -1,0 +1,54 @@
+import json
+import sys
+import subprocess
+import traceback
+
+
+def _remove_existing_tokens(token_ids: list, user_ocid: str, region: str) -> None:
+  """
+  Removes existing token if any for a user based on token id list, user_ocid and region
+  """
+  try:
+    for token in token_ids:
+      cmd = ["oci", "iam", "auth-token", "delete", "--auth-token-id", token.strip(), "--user-id", user_ocid, "--region", region, "--force"]
+      subprocess.check_output(cmd, text=True)
+  except subprocess.CalledProcessError as e:
+      raise ChildProcessError(f"Error in removing tokens for user: {user_ocid} {e.output}") from e
+
+
+def _regenerate_token(json_string:str) -> None:
+  """
+  Remove existing tokens and create a new API token. The json string is in the following format.
+  {"user_ocid: "id", "region": "oci_home_region", "token_ids":"t1,t2"}. Token ids are comma separated values.
+  """
+  input_json = json.loads(json_string)
+  user_ocid, region, token_ids = input_json["user_ocid"], input_json["region"], input_json["token_ids"].strip()
+  if token_ids:
+    token_ids = token_ids.split(",")
+    _remove_existing_tokens(token_ids, user_ocid, region)
+  try:
+    cmd = ["oci", "iam", "auth-token", "create", "--user-id", user_ocid, "--description", "autogen token", "--region", region]
+    output = subprocess.check_output(cmd, text=True)
+    result = json.loads(output)["data"]
+    output_data = {"token_id" : result["id"], "token" : result["token"]}
+    json_value = {"output" : json.dumps(output_data)}
+    print(json.dumps(json_value))
+  except subprocess.CalledProcessError as e:
+    raise ChildProcessError(f"Error in creating token for user: {user_ocid} {e.output}") from e
+
+
+if __name__ == "__main__":
+    input_json = sys.stdin.read()
+    if not input_json:
+      print("No Json provided", file=sys.stderr)
+      sys.exit(1)
+    err, trace = None, None
+    try:
+      _regenerate_token(input_json)
+      sys.exit(0)
+    except Exception as e:
+      trace = traceback.format_exc()
+      err = e
+    if trace:
+      print("Error:", err, trace, file=sys.stderr)
+      sys.exit(1)
--- a/datadog-oci-orm/metrics-setup/function_setup.tf
+++ b/datadog-oci-orm/metrics-setup/function_setup.tf
@@ -1,22 +1,55 @@
+data "oci_identity_users" "docker_registry_user" {
+  compartment_id = var.tenancy_ocid
+  name           = "DatadogAuthWriteUser"
+}
+
+data "oci_identity_auth_tokens" "docker_registry_user_tokens" {
+  count   = local.is_service_user_available ? 1 : 0
+  user_id = local.datadog_write_user_id
+}
+
+# Local variables to manage login and token value for the service account user.
+locals {
+  token_ids                 = [for token in data.oci_identity_auth_tokens.docker_registry_user_tokens : token.id]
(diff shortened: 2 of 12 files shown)


datadog-oci-orm/metrics-setup/function_setup.tf
```diff
@@ -1,37 +1,69 @@
...
-   }
+  }
+  triggers = {
+    always_run = "${timestamp()}"
+  }
 }
 
 ### Repository in the Container Image Registry for the container images underpinning the function
 resource "oci_artifacts_container_repository" "function_repo" {
+  count = local.is_service_user_available ? 1 : 0
```
sva91, line 66: Curious why this is dependent on presence of service user.

datadog-oci-orm/metrics-setup/function_setup.tf
```diff
@@ -1,37 +1,69 @@
...
+  count   = local.is_service_user_available ? 1 : 0
+  user_id = local.user_id
+}
+
+# Local variables to manage login and token value for the service account user.
+locals {
+  token_ids                 = [for token in data.oci_identity_auth_tokens.docker_registry_user_tokens : token.id]
+  is_service_user_available = length(data.oci_identity_users.docker_registry_user.users) == 1 ? true : false
+  user_login                = local.is_service_user_available ? data.oci_identity_users.docker_registry_user.users[0].name : ""
+  user_id                   = local.is_service_user_available ? data.oci_identity_users.docker_registry_user.users[0].id : ""
```
sva91, line 16: Nit: Might be good to rename this variable to be datadog_write_user_id. Same with user_login too. Currently, it's a little confusing which user we are talking about.

datadog-oci-orm/metrics-setup/auth_token.py
```diff
@@ -0,0 +1,54 @@
...
+  """
+  try:
+    for token in token_ids:
+      cmd = ["oci", "iam", "auth-token", "delete", "--auth-token-id", token.strip(), "--user-id", user_ocid, "--region", region, "--force"]
+      subprocess.check_output(cmd, text=True)
+  except subprocess.CalledProcessError as e:
+      raise ChildProcessError(f"Error in removing tokens for user: {user_ocid} {e.output}") from e
+
+
+def _create_token(json_string:str) -> None:
```
sva91, line 19: Nit: Since this is deleting and creating, maybe rename this to `_regenerate_token`.

datadog-oci-orm/metrics-setup/auth_token.py
```diff
@@ -0,0 +1,54 @@
...
+  user_ocid, region, token_ids = input_json["user_ocid"], input_json["region"], input_json["token_ids"].strip()
+  if token_ids:
+    token_ids = token_ids.split(",")
+    _remove_existing_tokens(token_ids, user_ocid, region)
+  try:
+    cmd = ["oci", "iam", "auth-token", "create", "--user-id", user_ocid, "--description", "autogen token", "--region", region]
+    output = subprocess.check_output(cmd, text=True)
+    result = json.loads(output)["data"]
+    output_data = {"token_id" : result["id"], "token" : result["token"]}
+    json_value = {"output" : json.dumps(output_data)}
```
sva91, line 34: Shouldn't we add atleast a 60 second sleep statement here if it takes 60-120 seconds for token to be available ? Else, I think we will pretty much always fail the first attempt.

datadog-oci-orm/policy-setup/policy.tf
```diff
@@ -29,55 +34,86 @@ resource "oci_identity_dynamic_group" "serviceconnector_group" {
...
+  defined_tags  = {}
+  freeform_tags = local.freeform_tags
+}
+
+resource "oci_identity_user" "write_permissions_user" {
+  #Required
+  compartment_id = var.tenancy_ocid
+  description    = "[DO NOT REMOVE] User for performing write based operations like docker image push"
+  name           = local.dd_auth_write_user
+  email          = "test@oci.com"
```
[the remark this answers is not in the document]
kanishktripathi: Here, an actual email actually gets sent to reset password (currently checking with security team on this). But since its a user with write access, having an email sent to our domain can be a concern

datadog-oci-orm/metrics-setup/auth_token.py
kanishktripathi: good point

datadog-oci-orm/metrics-setup/function_setup.tf
kanishktripathi: The idea is to not let anything to be created if we don't have the service user.

datadog-oci-orm/policy-setup/policy.tf
```diff
@@ -29,55 +34,86 @@ resource "oci_identity_dynamic_group" "serviceconnector_group" {
...
-  statements = ["Allow dynamic-group Default/${var.dynamic_group_name} to read metrics in tenancy",
-    "Allow dynamic-group Default/${var.dynamic_group_name} to use fn-function in tenancy",
-    "Allow dynamic-group Default/${var.dynamic_group_name} to use fn-invocation in tenancy",
-    "Allow group Default/${oci_identity_group.read_policy_group.name} to read all-resources in tenancy"
+  name           = local.datadog_metrics_policy
+  statements = ["Allow dynamic-group Default/${local.dynamic_group_name} to read metrics in tenancy",
+    "Allow dynamic-group Default/${local.dynamic_group_name} to use fn-function in tenancy",
+    "Allow dynamic-group Default/${local.dynamic_group_name} to use fn-invocation in tenancy",
+    "Allow group Default/${oci_identity_group.read_policy_group.name} to read all-resources in tenancy",
+    "Allow group Default/${oci_identity_group.write_user_group.name} to manage repos in tenancy where ANY {request.permission = 'REPOSITORY_READ', request.permission = 'REPOSITORY_UPDATE', request.permission = 'REPOSITORY_CREATE'}"
```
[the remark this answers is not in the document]
kanishktripathi: I tried without `REPOSITORY_CREATE ` and its not working. This is opposite of what OCI doc says

datadog-oci-orm/policy-setup/policy.tf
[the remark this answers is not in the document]
sva91: Hmm, it would be interesting if test@oci.com is an actual address used for something as well :)

sva91 (approved):

PR #40, opened 2025-02-12
[ECI-466] Metric Forwarding In Go

sva91:
## What
-Write Metrics Forwarder in Go

### Steps to build multi-architecture image

- Enable buildx : `docker buildx create --use`
- Build and push image : `docker buildx build --platform linux/amd64,linux/arm64 -t iad.ocir.io/iduhc9hzgn3o/metrics-forwarder-function/metrics-forwarder:0.0.3 . --no-cache --push`

## Why
- Onboarding improvements

## Testing
- Unit tests
- Staging [Logs](https://cloud.oracle.com/logging/log-groups/ocid1.loggroup.oc1.iad.amaaaaaaehwejaiawm226pvv74cvclkk6bvo3souwp2ajokq2nm2z526azza/logs/ocid1.log.oc1.iad.amaaaaaaehwejaiaq2ii55scta67hjdacn2gx7irx6nm75d27imv4ygpuhda/explore-log?region=us-ashburn-1)
- Using the same multi-platform image with function application of [ARM](https://cloud.oracle.com/functions/apps/ocid1.fnapp.oc1.iad.aaaaaaaajmefzoarr62vjhasd5rohuirhwvdr7judzfznykmx4eo442pqewq/fns?region=us-ashburn-1) and [x86](https://cloud.oracle.com/functions/apps/ocid1.fnapp.oc1.iad.aaaaaaaalvhvzuospksikxqpv5w7vg6orgalpo7534yh6np74s6fu5lbylja/fns?region=us-ashburn-1) architectures
- Invoked both the images through connector hub and it works fine


metrics-forwarder/forwarder.go
```diff
@@ -0,0 +1,85 @@
...
+	"bytes"
+	"compress/gzip"
+	"errors"
+	"fmt"
+	"log"
+	"net/http"
+	"os"
+)
+
+var httpClient = &http.Client{}
```
sva91, line 13: There may be some default settings that we can discuss, which might benefit here. ChatGPT's recommendation, but I am not sure about them:
```
var httpClient = &http.Client{
	Timeout: 5 * time.Second, // Reduce timeout for quick failures in serverless execution
	Transport: &http.Transport{
		MaxIdleConns:        10,  // Lower since function lifespan is short
		MaxIdleConnsPerHost: 5,   // Allow a few persistent connections per host
		IdleConnTimeout:     15 * time.Second, // Keep it low since function lifespan is short
		DisableKeepAlives:   false, // Keep HTTP Keep-Alive enabled for efficiency
		ForceAttemptHTTP2:   true,  // Optimize for HTTP/2 when available
		DialContext: (&net.Dialer{
			Timeout:   3 * time.Second, // Faster timeouts for transient functions
			KeepAlive: 30 * time.Second, // Useful if function container is reused
		}).DialContext,
		TLSHandshakeTimeout: 2 * time.Second,  // Fast TLS handshakes
		ExpectContinueTimeout: 1 * time.Second, // Reduce wait time for Expect: 100-continue
	},
}
```

metrics-forwarder/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM fnproject/go:1.23-dev AS builder
+WORKDIR /app
+COPY . . 
+
+# Enable Go modules & build with optimizations
+RUN go mod tidy && \
```
maxpyf, line 6: Do we need to support X86 as well?
I can't remember if OCI in Japan supports only ARM or only X86?

metrics-forwarder/forwarder.go
```diff
@@ -0,0 +1,85 @@
...
+
+type HTTPClient interface {
+	Do(req *http.Request) (*http.Response, error)
+}
+
+func sendMetricsToDatadog(metricsMessage []byte, client HTTPClient) error {
+	endpoint := os.Getenv("DD_INTAKE_HOST")
```
maxpyf, line 20: I see that this was the way in the Python version as well, so we can go with this for now...
Should we consider the DD API key a secret? 
From our perspective maybe we don't care (i.e. we charge more if someone sends more stuff to us) but from the customer's standpoint?
Anyway not a blocker for this PR by any means.

metrics-forwarder/forwarder.go
```diff
@@ -0,0 +1,85 @@
...
+	url := fmt.Sprintf("https://%s/api/v2/ocimetrics", endpoint)
+	apiHeaders := map[string]string{
+		"Content-Type": "application/json",
+		"DD-API-KEY":   apiKey,
+	}
+
+	if shouldCompressPayload() {
+		compressedPayload, err := compressPayload(metricsMessage)
+		if err != nil {
+			log.Printf("Error compressing payload: %v", err)
```
maxpyf, line 36: Also not a blocker for this version, but ultimately I think we should report this kind of error to us via the EvP endpoint, so that we can monitor when errors occur in the fleet and fix them proactively.

metrics-forwarder/forwarder.go
```diff
@@ -0,0 +1,85 @@
...
+		} else {
+			fmt.Printf("Uncompressed payload size=%d\n", len(metricsMessage))
+			metricsMessage = compressedPayload
+			apiHeaders["Content-Encoding"] = "Gzip"
+		}
+	}
+
+	// Send the metrics data
+	req, err := http.NewRequest("POST", url, bytes.NewBuffer(metricsMessage))
```
maxpyf, line 45: I'd do a `NewRequestWithContext` here, and choose a timeout context with a deadline of 4.5 min (a bit short of the 5 min deadline for Function runtime).
That way, things won't abruptly "disappear" if the function runs out of time.

metrics-forwarder/Dockerfile
rheei: Japan only supports x86 but I think we said with Kanishk's stack improvements we won't need to worry about it later on

metrics-forwarder/forwarder.go
maxpyf: Might I suggest using our own open-source client?
```"github.com/DataDog/datadog-api-client-go/v2/api/datadog"``` ?
We don't care about most of its features, only about the built-in configurations for things like this.

File: metrics-forwarder/Dockerfile (hunk 2 above)
Comment (maxpyf) [reply metrics-forwarder/Dockerfile:6]: well... we don't need to worry about it if we build for X86 and have multi arch images.
So I would say here we do need to build it as both.

File: metrics-forwarder/formatter.go [hunk 6]
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	if isDetailedLoggingEnabled() {
+		fmt.Println("Metric payload =", string(jsonData))
+	}
+	return jsonData, nil
+}
+
+func isDetailedLoggingEnabled() bool {
+	return os.Getenv("DETAILED_LOGGING_ENABLED") == "true"
```
rheei, line 77: I think in the previous forwarder we set the variable toLower before checking

metrics-forwarder/func.yaml
```diff
@@ -0,0 +1,12 @@
+schema_version: 20180708
+name: metrics-forwarder
+version: 0.0.13
+runtime: go
+build_image: fnproject/go:1.23-dev
+run_image: fnproject/go:1.23
+entrypoint: ./func
+memory: 256
```
maxpyf, line 8: is 256MB enough? just curious.

metrics-forwarder/func_test.go
```diff
@@ -0,0 +1,117 @@
...
+func (m MockFnContext) AppName() string                        { return "mock-app-name" }
+func (m MockFnContext) FnID() string                           { return "mock-fn-id" }
+func (m MockFnContext) FnName() string                         { return "mock-fn-name" }
+func (m MockFnContext) CallID() string                         { return "mock-call-id" }
+func (m MockFnContext) Config() map[string]string              { return nil }
+func (m MockFnContext) Header() http.Header                    { return nil }
+func (m MockFnContext) ContentType() string                    { return "" }
+func (m MockFnContext) TracingContextData() fdk.TracingContext { return nil }
+
+func TestMyHandler(t *testing.T) {
```
maxpyf, line 42: I like the tests. 
In the future I'd like us to really mock the HTTP responses from the DD metrics endpoint itself, so we can really ensure that the code behaves well in all cases. 
Not needed for this PR.

metrics-forwarder/func_test.go
```diff
@@ -0,0 +1,117 @@
...
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
```
rheei, line 13: nit pick: can delete this extra line in between external deps

metrics-forwarder/func_test.go
sva91: The tests in `metrics-forwarder/forwarder_test.go`, are those the ones you are looking for ?
I separated the tests for forwarder and rest of the functionality, for readability mostly, and also because I am mocking the forwarder function in the Handler tests here.

metrics-forwarder/forwarder.go
sva91: That makes complete sense, but I think that should be a separate PR with focus on "relaying information from forwarders to Datadog", so that we can think about holistically.

metrics-forwarder/formatter.go
sva91: Good catch

metrics-forwarder/response.go
```diff
@@ -0,0 +1,44 @@
...
+	"encoding/json"
+	"io"
+	"log"
+
+	fdk "github.com/fnproject/fdk-go"
+)
+
+// FnResponse defines the standard JSON response format
+type FnResponse struct {
+	Status  string `json:"status,omitempty"`
```
bengillmandd, line 13: ⛏️ should we omit empty for status?

metrics-forwarder/response.go
```diff
@@ -0,0 +1,44 @@
+package main
```
bengillmandd, line 1: I would also move this to a package in the `internal` directory

metrics-forwarder/formatter.go
```diff
@@ -0,0 +1,79 @@
...
+	Headers Header `json:"headers"`
+	Body    string `json:"body"`
+}
+
+type MetricsMessage struct {
+	Version string  `json:"version"`
+	Payload Payload `json:"payload"`
+}
+
+func generateMetricsMsg(ctx context.Context, serializedMetricData string) ([]byte, error) {
```
bengillmandd, line 35: ⛏️ Might be good to have a doc string here

metrics-forwarder/formatter.go
```diff
@@ -0,0 +1,79 @@
...
+	SourceFnName    string `json:"source_fn_name"`
+	SourceFnCallID  string `json:"source_fn_call_id"`
+}
+
+type Payload struct {
+	Headers Header `json:"headers"`
+	Body    string `json:"body"`
+}
+
+type MetricsMessage struct {
```
bengillmandd, line 30: Does this need to be an exported struct? maybe I'm missing something but it looks like it's only used in this file

metrics-forwarder/formatter.go
```diff
@@ -0,0 +1,79 @@
...
+type Header struct {
+	TenancyOCID     string `json:"tenancy_ocid"`
+	SourceFnAppOCID string `json:"source_fn_app_ocid"`
+	SourceFnAppName string `json:"source_fn_app_name"`
+	SourceFnOCID    string `json:"source_fn_ocid"`
+	SourceFnName    string `json:"source_fn_name"`
+	SourceFnCallID  string `json:"source_fn_call_id"`
+}
+
+type Payload struct {
```
bengillmandd, line 25: Does this need to be public?

metrics-forwarder/formatter.go
```diff
@@ -0,0 +1,79 @@
...
+	"fmt"
+	"os"
+	"strings"
+
+	fdk "github.com/fnproject/fdk-go"
+)
+
+var outputMessageVersion = "v1.0"
+
+type Header struct {
```
bengillmandd, line 16: Does this need to be public?

bengillmandd (commented):
overall lgtm just had some nitpicks

metrics-forwarder/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+WORKDIR /function
+
+# Copy only the compiled binary from the builder stage
+COPY --from=builder /app/func .
+
+# Set binary permissions
+RUN chmod +x /function/func
+
+# Run the application
+ENTRYPOINT ["./func"]
```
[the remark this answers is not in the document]
maxpyf: hold on... the entrypoint here is the main binary. So it can't really be "./main.go", it's not a .go file.

metrics-forwarder/formatter.go
maxpyf: This is just a small Function... it's good to be careful about these things, but there's only one package here. I doubt that this will ever make it into a bigger thing.
If it did, breaking it into multiple modules would be good, but I don't think we need to do this now...

metrics-forwarder/Dockerfile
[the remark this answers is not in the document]
bengillmandd: Sorry I meant here the entrypoint should be "main", referencing our "main" function in a "main.go" file which is the conventional entrypoint for any go module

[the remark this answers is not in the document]
maxpyf: that's different from the entrypoint in a container.
Here, it means "which binary is launched when the container starts". And that is definitely "./func" given how we called it. There is no rule, in Go or otherwise, that your binary must be called "main".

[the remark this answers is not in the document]
bengillmandd: Ah yes I meant "entrypoint to our go module". You're correct my comment is misplaced here. I will move it down to "func.go"

metrics-forwarder/func.go
```diff
@@ -0,0 +1,54 @@
+package main
```
bengillmandd, line 1: ⛏️  this file should be named "main.go" as this is the conventional entrypoint to go modules

metrics-forwarder/formatter.go
bengillmandd: I'd be ok to not put this into multiple files but we should have clean code. We don't want to end up with this:
image

maxpyf: there is a bit of room between this and what @sva91 wrote...

bengillmandd: haha I know I know. I'm just trying to help us have clean code but I agree, these are nit picks. I can refrain from adding these if they aren't useful

metrics-forwarder/forwarder.go
```diff
@@ -0,0 +1,85 @@
...
+	"fmt"
+	"log"
+	"net/http"
+	"os"
+	"strings"
+)
+
+var httpClient = &http.Client{}
+
+type HTTPClient interface {
```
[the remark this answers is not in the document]
bengillmandd: Yeah "caller" was a rushed suggestion before I had to go to an interview. The idea behind this naming convention is the interface is supposed to do something, thus the verb ('er') naming, and it makes interfaces easy to spot. This is another nit pick that I can refrain from adding in the future

metrics-forwarder/formatter.go
maxpyf: That's not what I am saying...  everything is a matter of context, aka "it depends".

For a project that is meant to evolve and become larger, these rules matter. I want to be able to read the code and understand what the intent is, and that is why they help. They help reduce the maintenance burden on other engineers, hence saving money for DD. 

For a small serverless function of this kind, unlikely to grow any bigger, it's probably okay not follow every naming convention in the book. We can easily follow what the intent is. The model exposed in these structs is unlikely to ever make it into a larger project.

In general, I am far more concerned about tests - because they help ensure that the code does what it says, and doesn't do what it shouldn't - that's the other way to make money for DD.
In this case, there are tests, but not around rate limits (429), which are not handled. Which is why I recommended using the Datadog API client over the vanilla Http client.

bengillmandd: understood

metrics-forwarder/forwarder.go
sva91: I have modified the client to use the one you recommended. Also, I set the Retry functionality ON for our client.

metrics-forwarder/forwarder.go
sva91: I think DD API Key should be considered a secret, but I don't fully follow what changes are you recommending based on that. Are you suggesting that this should not be a configuration for function application, but sourced from somewhere else (OCI vault etc.) ?

metrics-forwarder/forwarder.go
maxpyf: Perfect, very nice!

metrics-forwarder/forwarder.go
maxpyf: That is what I am saying, but of course it is not for this PR and maybe not even something we can/should do at all.
I am not recommending any changes today, but I think we should think about it as a team for the future.

sva91: Yea, I definitely agree that this may require some discussion, as well as some experimentation. It is certainly a good idea to add API-KEY and APP-KEY to the customer's vault during initial stack execution, and replicate it in all regions at that time. We can then use it everywhere we need this. 
I think this will be really nice for longer term. We just have to think about how to handle key rotations on expiration.

metrics-forwarder/Dockerfile
sva91: I have updated the image to build for both platforms. Also updated the PR description to have instructions on how to build such an image, and the testing I have done to ensure the same image works on function application of both architectures.

metrics-forwarder/Dockerfile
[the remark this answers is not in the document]
sva91: https://github.com/DataDog/oracle-cloud-integration/pull/40#discussion_r1959999280

metrics-forwarder/formatter.go
sva91: I do agree that the intent of this function is to be lightweight. To prove that, we can take a look at the equivalent metrics-forwarder in python code changes [history](https://github.com/DataDog/oracle-cloud-integration/commits/master/datadog-oci-orm/metrics-setup/metrics-function).
It has changed only once since it's creation, and that was also to add more debugging logs, i.e no change to existing code. 
But, I will try to accommodate some of your recommendations.

metrics-forwarder/forwarder.go
```diff
@@ -0,0 +1,85 @@
...
+
+	url := fmt.Sprintf("https://%s/api/v2/ocimetrics", endpoint)
+	apiHeaders := map[string]string{
+		"Content-Type": "application/json",
+		"DD-API-KEY":   apiKey,
+	}
+
+	if shouldCompressPayload() {
+		if compressedPayload, err := compressPayload(metricsMessage); err == nil {
+			fmt.Printf("Uncompressed payload size=%d\n", len(metricsMessage))
```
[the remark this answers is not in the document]
sva91: I am thinking that if we are generating one log statement per invocation, it should be ok. This is what we are doing in the python equivalent code. But I do see your point of value being added by the log statement in every invocation. I think it's valuable, since it gives us an idea on the size of payload being forwarded to Datadog, which gives us some indication of throughput.
Let me know what you think.

metrics-forwarder/response.go
sva91: I thought it would make sense to omit all empty fields. Any reason we shouldn't ?

metrics-forwarder/forwarder.go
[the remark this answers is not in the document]
sva91: I have modified the code to have the interface implement multiple methods now. The link you shared refers to interfaces with 1 method implementation. I do think the interface name of "APIClient" is more readable, and makes sense.

metrics-forwarder/formatter.go
sva91: updated

metrics-forwarder/formatter.go
sva91: updated

metrics-forwarder/forwarder.go
```diff
@@ -0,0 +1,85 @@
+package main
```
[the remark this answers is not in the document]
sva91: @bengillmandd I have made the changes. I kept it very simple initially, thinking that a stable lightweight function doesn't require a lot of maintenance. But I anyways did it for some Go practice, since structuring doesn't really hurt in any way. Thanks.

metrics-forwarder/response.go
bengillmandd: Ah I was just curious if we would ever expect there to be a function response without a Status. Not a big deal in this case

metrics-forwarder/forwarder.go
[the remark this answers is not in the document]
bengillmandd: Ah yes this is fine now. i think before there were print lines for uncompressed and compressed size. Customers have to pay (I'm not sure how much) for function logs, this is the only reason I mentioned trying to limit them

bengillmandd (approved):

metrics-forwarder/internal/client/forwarder.go
```diff
@@ -0,0 +1,59 @@
...
+// request to the Datadog API and checks the response status code.
+//
+// Parameters:
+//   - client: A DatadogClient instance containing the API key and HTTP client.
+//   - metricsMessage: A byte slice containing the metrics data to be sent.
+//
+// Returns:
+//   - error: An error if the request preparation or API call fails, or if the
+//     response status code indicates a failure.
+func SendMetricsToDatadog(client DatadogClient, metricsMessage []byte) error {
```
kanishktripathi, line 26: can this be a method of  `DatadogClient` ?

metrics-forwarder/internal/client/client.go
```diff
@@ -0,0 +1,32 @@
...
+
+	datadog "github.com/DataDog/datadog-api-client-go/v2/api/datadog"
+)
+
+type apiClient interface {
+	CallAPI(req *http.Request) (*http.Response, error)
+	PrepareRequest(ctx context.Context, path string, method string, postBody interface{}, headerParams map[string]string, queryParams url.Values, formParams url.Values, fileName *datadog.FormFile) (*http.Request, error)
+}
+
+type DatadogClient struct {
```
kanishktripathi, line 16: not immediate but this part is a re-usable code when creating log forwarding client as well

metrics-forwarder/Dockerfile
kanishktripathi: Agreed, the multi-arch image build is managed outside the docker file using `buildx` so that will be a separate step

metrics-forwarder/internal/client/client.go
```diff
@@ -0,0 +1,32 @@
...
+	Client apiClient
+	site   string
+	apiKey string
+}
+
+func CreateDatadogClient(site, apiKey string) DatadogClient {
+	configuration := datadog.NewConfiguration()
+	configuration.RetryConfiguration.EnableRetry = true
```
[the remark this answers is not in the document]
sva91: Default setting for these values are as follows ([Link](https://github.com/DataDog/datadog-api-client-go/blob/c974d03fa124fe3330931b9a2d94f9a1a9e89030/api/datadog/configuration.go#L406)):
```
RetryConfiguration: RetryConfiguration{
	EnableRetry:       false,
	BackOffMultiplier: 2,
	BackOffBase:       2,
	HTTPRetryTimeout:  60 * time.Second,
	MaxRetries:        3,
}
```
If we want to change them, we can override, but they looked fine to me as is.

metrics-forwarder/internal/client/forwarder.go
```diff
@@ -0,0 +1,59 @@
...
+	if err != nil {
+		return err
+	}
+
+	resp, err := client.Client.CallAPI(req)
+	if err != nil {
+		return err
+	}
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		log.Printf("Error: Received non-200 response from Datadog: %d", resp.StatusCode)
```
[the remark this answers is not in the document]
sva91: Great question. I experimented with both. Here is how these different log packages are treated in OCI.

- Log package : These are written to STDERR by default. You will see `"src" : "STDERR"` in the log json.
- fmt package : These are written to STDOUT by default. You will see `"src" : "STDOUT"` in the log json.

If a log is an error, I used the log package, else I used the fmt package.

metrics-forwarder/internal/handler/response.go
```diff
@@ -0,0 +1,44 @@
...
+		log.Printf("Error marshalling response: %v", jsonErr)
+		fdk.WriteStatus(out, 500)
+		out.Write([]byte(`{"status":"error","error":"Internal server error"}`))
+		return
+	}
+
+	if status == "success" {
+		fdk.WriteStatus(out, 200)
+	} else {
+		fdk.WriteStatus(out, 500)
```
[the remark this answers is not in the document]
sva91: Hmm, I am considering the response from function invocations as a pass or fail only. The status code itself will be printed in OCI logs with the status that we got back from Datadog.

This particular status code becomes relevant if the function itself is placed behind an API Gateway, and then invoked through an API call. Now, for the client of the function itself, if an internal API call is failing, should we mask that or forward that as is could be debated.

Let me know if you want to have a discussion about this.

metrics-forwarder/internal/handler/handler.go
```diff
@@ -0,0 +1,83 @@
...
+func getSerializedMetricData(rawMetrics io.Reader) (string, error) {
+	buf := new(bytes.Buffer)
+	_, err := buf.ReadFrom(rawMetrics)
+	if err != nil {
+		return "", err
+	}
+	return buf.String(), nil
+}
+
+func readEnvVars() (string, string, string, error) {
```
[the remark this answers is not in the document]
sva91: Yes, good suggestion. I agree returning tuples of same type can lead to unintended bugs.

metrics-forwarder/internal/client/forwarder.go
sva91: I am not sure I fully understand the intent behind making this a function on struct, instead of passing the struct. This way, it helps me test the send function by creating a mock client.

metrics-forwarder/internal/handler/handler.go
```diff
@@ -0,0 +1,83 @@
...
+	"errors"
+	"io"
+	"log"
+	"os"
+
+	"metrics-forwarder/internal/client"
+	"metrics-forwarder/internal/formatter"
+)
+
+var sendMetricsFunc = client.SendMetricsToDatadog
```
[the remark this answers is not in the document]
sva91: I am not sure I fully understand this comment. I am happy to chat more about this.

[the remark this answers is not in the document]
sva91: I have added the cache in the client package itself. Please take a look : https://github.com/DataDog/oracle-cloud-integration/pull/40/commits/0182be1b689b773936b8a93e5759888d1aff98bf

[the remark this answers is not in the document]
maxpyf: I think this is fine, though you could also simply pass it as an argument to `MyHandler`.

metrics-forwarder/internal/handler/handler.go
```diff
@@ -0,0 +1,83 @@
...
+	// 1. Read metrics data
+	serializedMetricData, err := getSerializedMetricData(in)
+	if err != nil {
+		log.Printf("Error reading metric data: %v", err)
+		writeResponse(out, "error", "", err)
+		return
+	}
+
+	// 2. Read tenancy OCID, site and datadog API key
+	client, tenancyOCID, err := newDatadogClientWithTenancy()
```
maxpyf, line 39: the variable `client` shadows the internal module `client`. Maybe call the variable `ddClient` or something a bit different (not necessary for merging).

maxpyf (approved):
one minor comment, approved!
nice work.

metrics-forwarder/internal/handler/handler.go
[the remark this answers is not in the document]
maxpyf: You could return the right signature with a closure though.
It is just like the HTTPHandler pattern / middleware pattern in net/http.

In any case, it is not important, this works fine as is.

PR #41, opened 2025-02-19
Log Forwarder In Go

sva91:


logs-forwarder/internal/client/forwarder.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"log"
+	"net/http"
+	"os"
+	"strings"
+
+	datadog "github.com/DataDog/datadog-api-client-go/v2/api/datadog"
```
rheei, line 11: nit pick: you probably don't need to name this datadog if the package is already called datadog

logs-forwarder/internal/formatter/formatter.go
```diff
@@ -0,0 +1,108 @@
...
+	"os"
+	"strings"
+)
+
+const DD_SERVICE = "oci"
+const DD_SOURCE = "oci.logs"
+
+// ProcessedLog represents the transformed log format.
+type logPayload struct {
+	Source    string                 `json:"source,omitempty"`
```
rheei, line 14: I remember discussing with you that having both "source" and "ddsource" fields is kind of misleading, I'm not sure what we would call this but potentially we could think of a different attribute name to map it to ? If anyone else has thoughts

logs-forwarder/internal/handler/handler.go
```diff
@@ -0,0 +1,124 @@
...
+	if err != nil {
+		log.Println(err)
+		writeResponse(out, "error", "", err)
+		return
+	}
+
+	batchSize := getBatchSize()
+	fmt.Printf("Received %d logs to process with a batch size of %d\n", len(logs), batchSize)
+	for i := 0; i < len(logs); i += batchSize {
+		end := i + batchSize
```
bengillmandd, line 53: Is there a maximum size in bytes for these batches? I'm wondering if we might run into a similar problem like we had with metrics

logs-forwarder/internal/handler/handler.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+	return logs, nil
+}
+
+func processLogs(logs []map[string]interface{}, site, apiKey string) error {
+	logsMsg, err := formatter.GenerateLogsMsg(logs)
+	if err != nil {
+		return err
+	}
+	err = sendLogsFunc(client.CreateDatadogClient(site, apiKey), logsMsg)
```
bengillmandd, line 97: Do we need to create a client for each batch? Maybe we can save time by using the same client?

logs-forwarder/internal/formatter/formatter_test.go
```diff
@@ -0,0 +1,183 @@
...
+			var logPayloads []logPayload
+			err = json.Unmarshal(result, &logPayloads)
+			assert.NoError(t, err)
+
+			assert.Equal(t, tt.expected, logPayloads)
+		})
+	}
+}
+
+func TestRedactField(t *testing.T) {
```
bengillmandd, line 116: nitpick: technically this tests code that's in redactor.go, should be in its own test file

bengillmandd (commented):
overall lgtm just had a few comments

logs-forwarder/internal/formatter/utils.go
```diff
@@ -0,0 +1,36 @@
...
+	"data.request.headers.Authorization":                  {},
+	"data.request.headers.authorization":                  {},
+	"data.request.headers.X-OCI-LB-PrivateAccessMetadata": {},
+	"data.request.headers.opc-principal":                  {},
+}
+
+// Loads the exclude list from an environment variable
+// EXCLUDE Env Variable Example: '{ "data.identity": [ "credentials"], "data.request.headers": [ "authorization", "Authorization", "X-OCI-LB-PrivateAccessMetadata", "opc-principal" ] }'
+func getExcludeList() (map[string]struct{}, error) {
+	excludeJSON := os.Getenv("EXCLUDE")
```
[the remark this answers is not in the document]
sva91: That's right. The fields that we decided should always be redacted (check out utils.go), will always be redacted. This is an addition, which can be used if needed, to redact any additional fields. 
I followed the same pattern as described by OCI for achieving that [here](https://docs.oracle.com/en/learn/ocilogs-siem-func/index.html#objectives)

logs-forwarder/internal/formatter/formatter.go
sva91: Yes, that makes sense. I think while playing around, from what I understood:

- DDSOURCE: A field which is required by Logs Intake to build custom pipelines.
- Source: A field which we are passing forward as is from OCI logs. It's similar to other fields like timestamp, type etc.

The confusion with the above 2 fields is how they are displayed on the Logs UI.
Would love to hear feedback from others if this looks ok to them, or if we should change something here. One option could be to rename source to `OCISource`.

Screenshot 2025-02-20 at 10 47 04 AM

logs-forwarder/internal/handler/handler.go
sva91: There is a 5 MB limit, but the bytes for function invocations should be controlled from the target settings in Connector Hub [here](https://cloud.oracle.com/connector-hub/service-connector-edit/ocid1.serviceconnector.oc1.iad.amaaaaaaehwejaiap3q7x6rw5s37ycyz4yqygrgr4cl57674op5eviezbfpa?region=us-ashburn-1)

logs-forwarder/internal/handler/handler.go
sva91: This is a great point. Thanks for catching that.

logs-forwarder/internal/formatter/formatter.go
rheei: I was also thinking something along the lines of "OCISource", this field is kinda flaky so imo we should name it something different so customers don't get confused

bengillmandd (approved):

logs-forwarder/internal/formatter/formatter.go
maxpyf: I agree we should have 2 different names - anything like this will confuse customers. Renaming source to OCISource or whatever other name PM finds acceptable is a good solution.

I am curious what things look like in other clouds? Can we imitate what they do?
In this example, is "DDSOURCE" meant to be "oci.functions"?

rheei: the "ddsource" field is the "entry point" for log pipelines. It must be unique and the pattern we are following (along with gcp and azure) is that the dd source is the cloud + the cloud service where the log is coming from. So for example, the VCN flow logs ddsource field is oci.vcn, and that way these logs are correctly processed through their respective pipeline

logs-forwarder/internal/handler/handler.go
maxpyf: One issue here is that our current Go SDK does not allow setting these batch limits programmatically. I created ECI-503 for this.

logs-forwarder/internal/formatter/formatter.go
```diff
@@ -0,0 +1,108 @@
...
+		return nil, err
+	}
+	tags := getTags()
+	logPayloads := make([]logPayload, len(logs))
+	for i, log := range logs {
+		applyRedaction(log, excludeSet)
+		logPayload := formatLog(log)
+		logPayload.DDTags = tags
+		logPayloads[i] = logPayload
+	}
```
sva91, line 43: @maxpyf , @kanishktripathi Do you folks think it might be beneficial to use go routines here ? There is no network call.

maxpyf (approved):
Approved!

The only potential issue with the implementation is the "batch" decoding:
a whole batch is decoded, converted to a slice of maps, redacted, re-encoded, and then pushed to DD.

This might be good enough for our needs, I don't think it blocks us for now.
But we might need to consider a "streaming" implementation where we "pipeline" the messages through the transformations and eventually to DD, just to save memory.

We might just need to bump the memory.

Please just fix the vulns reported by the bot.

logs-forwarder/internal/formatter/formatter.go
maxpyf: I think it's always beneficial to "pipeline" in Go, and let the runtime manage go routines and concurrency.
It might make things a bit faster.

sva91: Yes, that makes sense. I will modify this.

sva91:
I will be addressing the remaining open comments in a new PR.

PR #42, opened 2025-02-25
[ECI-492] Create Installation stack

sva91:
## What

- Initial Onboarding stack
- Creates compartment, vcn with nat gateway and service gateway
- Creates function app, and functions
- Creates auth user, TLS OCI API Key pairs, dynamic groups and policies
- DDAuthUser gets the email of the Admin running the stack

## Why

- Improving Onboarding experience

## Testing

- https://cloud.oracle.com/resourcemanager/stacks/ocid1.ormstack.oc1.iad.amaaaaaaehwejaia7daa4dpxatwpqt3rqon4fnxlfjjfyuvq7xzr5b5n467a/jobs?region=us-ashburn-1

### DIFF
--- a/datadog-integration/modules/functions/main.tf
+++ b/datadog-integration/modules/functions/main.tf
@@ -13,7 +13,7 @@ resource "oci_functions_function" "logs_function" {
   count           = var.logs_image_tag != "" ? 1 : 0
   application_id  = oci_functions_application.dd_function_app.id
   display_name    = "dd-logs-forwarder"
-  memory_in_mbs   = "256"
+  memory_in_mbs   = "1024"
   freeform_tags   = var.tags
   image           = local.logs_image_path
 }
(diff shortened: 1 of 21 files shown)


datadog-integration/modules/functions/main.tf
```diff
@@ -0,0 +1,28 @@
...
+    var.subnet_id,
+  ]
+  config         = local.config
+}
+
+resource "oci_functions_function" "logs_function" {
+  count           = var.logs_image_tag != "" ? 1 : 0
+  application_id  = oci_functions_application.dd_function_app.id
+  display_name    = "dd-logs-forwarder"
+  memory_in_mbs   = "256"
```
maxpyf, line 16: should we bump these up? 
not relevant for milestone 1 but in general I worry that we are limiting ourselves a bit.

maxpyf (approved):
looks good.

One question: in which compartment is this stack created?
For monitoring that it works from our backend, we may need to call some APIs (ListStack,  GetJobLogs).  I am thinking we should add these permissions now to our DDAuthUser.
The resource type is `orm-jobs` and `orm-stacks`, and we would need `read` to get the log and errors. I think that will greatly help us down the line.

datadog-integration/modules/functions/main.tf
sva91: Yea, I think we should. Does 1024 suffice, or do you think we should bump it even more ?

PR #46, opened 2025-03-12
ECI-529 Do not do docker login on every stack run

kanishktripathi:
**what:**

* The docker login re-creates a service user auth token and does a login which takes at least 2-3 minutes.
* This can be skipped if not creating a function image and only refreshing connector hub compartments.


sva91 (approved):
LGTM

PR #47, opened 2025-03-19
[ECI-518] Read API Keys from vault

sva91:
## What
- Refactor client package to be shared among logs-forwarder and metrics-forwarder
- Read API Key from Secret OCID config variable
- Update Dockerfiles to build images correctly
- Adding policies for functions to read secrets
- Remove DD_COMPRESS variable, as it doesn't make much sense (we should always compress)

## Building and Pushing Image to OCI

1. cd into `datadog-functions` directory
2. Example command: `docker buildx build --platform linux/amd64,linux/arm64 -t iad.ocir.io/iduhc9hzgn3o/sva-datadog-functions/logs-forwarder:0.0.12 . --no-cache -f logs-forwarder/Dockerfile --push`

## Why
- Securing API Keys in Vault, so that they aren't visible in function app configuration

## Testing
- Deployed the images in OCI and update function app to use these images. [Link](https://cloud.oracle.com/functions/apps/ocid1.fnapp.oc1.iad.aaaaaaaajmefzoarr62vjhasd5rohuirhwvdr7judzfznykmx4eo442pqewq/fns?region=us-ashburn-1)
- Invoked the functions through CLI, as well as connector hub to ensure expected functionality
- Deployed the function in FRA (Frankfurt) region as well, while having the vault in us-ashburn-1. Tested cross region access for vault
- Command for CLI Execution : `oci fn function invoke --function-id ocid1.fnfunc.oc1.iad.aaaaaaaaabzbzoulordjcyolsdkn5z5im6xyvkyxjwdixhq7vs6si4mgsdxa --body '{"key":"value"}' --file -`


datadog-integration/modules/auth/locals.tf
```diff
@@ -1,8 +1,9 @@
 locals {
   user_group_name  = "${var.user_name}-admin"
   user_policy_name = "${var.user_name}-policy"
-  dg_name          = "dd-sch-dg"
-  dg_policy_name   = "${local.dg_name}-policy"
+  dg_sch_name      = "dd-connectorhubs"
```
maxpyf, line 4: This appears to be a dynamic group name.
I know this is verbose and redundant, but maybe `dd-dynamicgroup-connectorhubs` or maybe just `dd-dyngroup-xxx` ?
I wasn't sure if these were actual connectorhubs with this name.

datadog-integration/modules/auth/main.tf
```diff
@@ -69,12 +79,12 @@ resource "oci_identity_policy" "dg_policy" {
   description    = "[DO NOT REMOVE] Policy to have any connector hub read from eligible sources and write to a target function"
   name           = local.dg_policy_name
   statements = concat(
-    var.forward_logs ? ["Allow dynamic-group Default/${local.dg_name} to read log-content in tenancy"] : [],
-    var.forward_metrics ? ["Allow dynamic-group Default/${local.dg_name} to read metrics in tenancy"] : [],
+    var.forward_logs ? ["Allow dynamic-group Default/${local.dg_sch_name} to read log-content in tenancy"] : [],
```
maxpyf, line 82: Nice. Our "tenancy-wide" permissions are read-only, and our "management" permissions are scoped to our compartment. We will need to make this very clear in the documentation.

internal/common/client.go
```diff
@@ -0,0 +1,107 @@
...
+//	error - An error if the message could not be sent or if the API key could not be refreshed.
+func (client DatadogClient) SendMessageToDatadog(ctx context.Context, message []byte, url string) error {
+	status, err := client.sendMessage(ctx, message, url)
+	if err != nil && status == http.StatusForbidden {
+		// Attempt to fetch the API key again in case it has been rotated
+		err = client.refreshAPIKey(ctx)
+		if err != nil {
+			return err
+		}
+		_, err = client.sendMessage(ctx, message, url)
```
maxpyf, line 49:
```suggestion
		return client.sendMessage(ctx, message, url)
```

internal/common/client.go
```diff
@@ -0,0 +1,107 @@
...
+	return err
+}
+
+func (client DatadogClient) sendMessage(ctx context.Context, message []byte, url string) (int, error) {
+	apiHeaders := map[string]string{
+		"Content-Encoding": "gzip",
+		"Content-Type":     "application/json",
+		"DD-API-KEY":       client.ApiKey,
+	}
+	fmt.Printf("Uncompressed payload size=%d\n", len(message))
```
maxpyf, line 61: Does this go to the invocation logs for this function?

internal/common/client.go
```diff
@@ -0,0 +1,107 @@
...
+	configuration := datadog.NewConfiguration()
+	configuration.RetryConfiguration.EnableRetry = true
+	client := datadog.NewAPIClient(configuration)
+
+	cache = &DatadogClient{
+		Client:      client,
+		secretOCID:  secretOCID,
+		vaultRegion: homeRegion,
+	}
+	apiKey, err := cache.fetchAPIKeyFromVault(context.Background())
```
maxpyf, line 101: Since the function will timeout after 5 minutes, I would use a `WithTimeout` context, with a timeout of e.g. 3 minutes. That way, if it does timeout, the function will gracefully shut down, and we will know exactly why.

internal/common/vault.go
```diff
@@ -0,0 +1,65 @@
...
+	"encoding/base64"
+	"errors"
+	"fmt"
+
+	"github.com/oracle/oci-go-sdk/v65/common"
+	"github.com/oracle/oci-go-sdk/v65/common/auth"
+	"github.com/oracle/oci-go-sdk/v65/secrets"
+)
+
+func (client DatadogClient) refreshAPIKey(ctx context.Context) error {
```
maxpyf, line 14: You need a pointer receiver here, unless I am missing something.
Go is "by value", so this method is operating on a copy of client.
The assignment `client.ApiKey = apiKey` will not survive this method call and that does not seem like the intent.

logs-forwarder/internal/handler/handler.go
```diff
@@ -123,3 +115,12 @@ func getBatchSize() int {
 	}
 	return batchSize
 }
+
```
maxpyf, line 118: I would consider making this function another constructor in the client module. It feels like it belongs there.

maxpyf (approved):
Looks good overall, nice job!

Please just fix that pointer receiver issue in the client and you are good to go from my side.
You might also want to update the libs so that the bot doesn't complain about vulns.

datadog-integration/modules/auth/locals.tf
```diff
@@ -1,8 +1,9 @@
...
+  dg_fn_name       = "dd-functions"
```
kanishktripathi, line 5: same with this variable name, it might be verbose, but will nice to name these clearly

internal/go.mod
```diff
@@ -0,0 +1,26 @@
...
+
+go 1.23.4
+
+require (
+	github.com/DataDog/datadog-api-client-go/v2 v2.36.1
+	github.com/oracle/oci-go-sdk/v65 v65.87.0
+	github.com/stretchr/testify v1.8.4
+)
+
+require (
```
kanishktripathi, line 11: Can this be in a separate directory like this with a single directory for all go code
```
forwarder
  - common
  - metrics
  - logs
```
Unless we plan to use this directory outside the forwarder code

sva91: That's a good idea, but I don't think we should call the root folder forwarder. We could potentially use datadog-functions as the root folder name, not limiting the common package to only forwarder style functions. I will make this change.

internal/common/client.go
sva91: Yes, they do. We can see these logs [here](https://cloud.oracle.com/logging/log-groups/ocid1.loggroup.oc1.iad.amaaaaaaehwejaiawm226pvv74cvclkk6bvo3souwp2ajokq2nm2z526azza/logs/ocid1.log.oc1.iad.amaaaaaaehwejaiaq2ii55scta67hjdacn2gx7irx6nm75d27imv4ygpuhda?region=us-ashburn-1)

internal/common/vault.go
sva91: Great catch. Thanks.

datadog-functions/internal/client/client.go
```diff
@@ -0,0 +1,130 @@
...
+//	ctx - The context for the request.
+//	message - The message to be sent as a byte slice.
+//	url - The URL to which the message should be sent.
+//
+// Returns:
+//
+//	error - An error if the message could not be sent or if the API key could not be refreshed.
+func (client *DatadogClient) SendMessageToDatadog(ctx context.Context, message []byte, url string) error {
+	ctx, cancel := context.WithTimeout(ctx, 3*time.Minute)
+	defer cancel() // releases resources if sendMessage or refreshAPIKey completes before timeout elapses
```
sva91, line 44: @maxpyf in line with the comment [here](https://github.com/DataDog/oracle-cloud-integration/pull/47#discussion_r2005666781),I also added the WithTimeout context when making API Calls to Datadog.
Please take a look and let me know if this looks alright.

maxpyf: looks good!

PR #48, opened 2025-03-28
ECI-549 Reduce smaller base image and use the base image from the repo

kanishktripathi:
**what:**
* Use the base image from the datadog ECR repository which mirrors the added docker images
* Build from the golang image and use scratch as the base image for the Go binary.
* Rename client module from `internal` to `lib. Namely to distinguish from the internal directory in each forwarder.
* Moved the dockerfiles outside the modules to copy with absolute path names from the docker path
* Added a README for build instructions.

**why:**
* Reduce the final image size. Final image size is around 7.5Mb
* Some cosmetic changes in the naming. 
* Created instructions on how to build images locally and from which path.

**testing:**

* Tested by docker build and then pushing the image for [these functions](https://cloud.oracle.com/functions/apps/ocid1.fnapp.oc1.iad.aaaaaaaajmefzoarr62vjhasd5rohuirhwvdr7judzfznykmx4eo442pqewq/fns?region=us-ashburn-1&cloudshell=true&bdcstate=default)

Image versions.

Log forwarder: `iad.ocir.io/iduhc9hzgn3o/sva-datadog-functions/logs-forwarder:0.0.17`
Metrics forwarder: `iad.ocir.io/iduhc9hzgn3o/sva-datadog-functions/metrics-forwarder:0.0.7`

Log lines for metrics and logs:

Screenshot 2025-03-28 at 2 31 41 PM

Screenshot 2025-03-28 at 2 31 05 PM


datadog-functions/Dockerfile-logs
```diff
@@ -0,0 +1,20 @@
+FROM registry.ddbuild.io/images/mirror/golang:1.24.0 AS builder
+WORKDIR /app
+COPY lib/ lib/
+COPY logs-forwarder/ logs-forwarder/
+
+# Enable Go modules & build with optimizations
+WORKDIR /app/logs-forwarder
+RUN go mod tidy  && \
```
maxpyf, line 8: what happened to `GOARCH=$(go env GOARCH)`?
I thought we agreed we needed multi-arch images to serve regions where ARM is not available.

kanishktripathi: Added this back. Made it dependent on the target platform. Multi arch images are built using `docker buildx` and the actual build is based on the platform parameter passed to it.

Example:
`docker buildx build --tag <img_tag_to_registry> --output type=registry --platform linux/amd64,linux/arm64   -f Dockerfile-metrics .`

PR #57, opened 2025-05-14
ECI-630 Add subscribed regions to be added as a stack

kanishktripathi:
**what:**
* Spin additional stacks for Functions and network for each region.
* In case of a destroy, run destroy on the stacks and delete the stack.
* Each stack is created in its own region and not on home region because each region has a limit of 5 concurrent stack jobs to run.

**why:**
* To spin up functions and network resources for metrics and log forwarding in all regions.

**testing:**
On **datadog84**, [see here](https://cloud.oracle.com/resourcemanager/stacks/ocid1.ormstack.oc1.iad.amaaaaaaehwejaia33vhursf2lq32klf5kccenvnwdk2qz7bywgp3imwqipq?region=us-ashburn-1)


datadog-integration/main.tf
```diff
@@ -61,3 +42,56 @@ module "integration" {
...
+    working_dir = path.module
+    when        = create
+    command     = <<EOT
+
+    echo "Checking any existing stacks in the compartment and destroy them...."
+
+    chmod +x ${path.module}/delete_stack.sh; ${path.module}/delete_stack.sh ${self.input.compartment} ${each.key}
+
+    STACK_ID=$(oci resource-manager stack create --compartment-id ${self.input.compartment} --display-name datadog-regional-stack-${each.key} \
+      --config-source ${path.module}/modules/regional-stacks/dd_regional_stack.zip  --variables '{"tenancy_ocid": "${var.tenancy_ocid}", "region": "${each.key}", \
```
sva91, line 72: for display name of stack, I think we can just use `datadog-regional-stack` probably, since the stack itself is being deployed in the region of interest.

datadog-integration/delete_stack.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+COMPARTMENT=$1
+REGION=$2
+
+STACK_IDS=($(oci --region "$REGION" resource-manager stack list --compartment-id $COMPARTMENT --raw-output | jq -r '.data[]."id"'))
```
sva91, line 6: we can probably use [display name](https://docs.oracle.com/en-us/iaas/tools/oci-cli/3.56.0/oci_cli_docs/cmdref/resource-manager/stack/list.html#cmdoption-display-name) filter here to best guess stack created by this tf, and remove that only.

datadog-integration/locals.tf
```diff
@@ -11,16 +11,26 @@ locals {
...
+
+}
+
+
+# Variables for regions
+locals {
+  subscribed_regions = [for region in data.oci_identity_region_subscriptions.subscribed_regions.region_subscriptions : region]
+
+  # All supported regions list based on docker image existence
+  supported_regions      = data.external.supported_regions
```
sva91, line 24: Is this variable `supported_regions` used anywhere ?

datadog-integration/locals.tf
```diff
@@ -11,16 +11,26 @@ locals {
...
-    for region in data.oci_identity_region_subscriptions.subscribed_regions.region_subscriptions :
+    for region in local.subscribed_regions :
     region.region_name => region
   })
 
-  subscribed_regions_set = toset(local.subscribed_regions_list)
 
-  is_current_region_home_region = (var.region == local.home_region_name)
+
+  subscribed_regions_set = toset(local.supported_regions_list)
```
sva91, line 35: maybe rename this to be `supported_subscribed_regions_set` to prevent confusion with subscribed regions.

datadog-integration/regional_stacks.tf
```diff
@@ -0,0 +1,88 @@
...
+    
+    if [[ "$VALUE" != "" ]]; then
+      echo "The region ${each.key} is not supported.....exit"
+      exit 0
+    fi
+
+    echo "Checking any existing stacks in the compartment...."
+    
+    # The name of the stack to be created. Combined with the stack_digest to make it unique to this stack
+    STACK_NAME="datadog-regional-stack-${terraform_data.stack_digest.id}"
```
sva91, line 39: Finding it hard to understand this. So, we create a new regional stack every time a new apply job is executed on onboarding stack. If we do that, we create a new stack with the latest code in regional_stacks.tf. Will we end up creating vcns and functions every time ? 
I guess I am not understanding the benefit of stack_digest. Let me know if it would be quicker to chat and clarify.

datadog-integration/delete_stack.sh
kanishktripathi: done

datadog-integration/regional_stacks.tf
sva91: spoke about this issue.

sva91 (approved):

PR #58, opened 2025-05-20
hide variables and remove fingerprint

sva91:
- Hide Datadog Site, App Key and API Key variables in stack
- Don't pass fingerprint in create API
Screenshot 2025-05-21 at 1 24 14 PM
Screenshot 2025-05-21 at 1 23 54 PM

### DIFF
--- a/datadog-integration/schema.yaml
+++ b/datadog-integration/schema.yaml
@@ -26,9 +26,9 @@ variableGroups:
 variables:
   # Advanced Options
   allow_datadog_env_changes:
-    title: Allow Datadog Environment Variable Changes
+    title: Override Datadog Variables
     type: boolean
-    description: Enable or disable the ability to modify Datadog environment variables
+    description: Click to view/modify pre-set Stack variables
     required: false
     default: false
 
--- a/datadog-integration/schema.yaml
+++ b/datadog-integration/schema.yaml
@@ -12,31 +12,16 @@ variableGroups:
       - ${region}
       - ${compartment_ocid}
       - ${current_user_ocid}
+    visible: false
   - title: "Datadog Environment"
     variables:
       - ${datadog_app_key}
       - ${datadog_api_key}
       - ${datadog_site}
-  - title: "Tenancy Options"
-    variables:
-      - ${metrics_image_tag}
-      - ${logs_image_tag}
+    visible: false
 
 variables:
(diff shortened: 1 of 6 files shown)


natwilkinson (approved):

datadog-integration/schema.yaml
```diff
@@ -12,31 +12,16 @@ variableGroups:
...
 
-  compartment_ocid:
-    visible: false
-  
-  current_user_ocid:
-    visible: false
-
-# Datadog Environment
+variables:
+  # Datadog Environment
```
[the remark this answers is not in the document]
sva91: I think you mean for the DATADOG_SITE. Is the scenario you are referring to is when an org migrates from one datacenter to another ? 
I think it might be beneficial to not make this an enum, to avoid any stack changes when Datadog expands to newer datacenters (like AP2).

datadog-integration/schema.yaml
```diff
@@ -12,31 +12,16 @@ variableGroups:
       - ${region}
       - ${compartment_ocid}
       - ${current_user_ocid}
+    visible: false
```
maxpyf, line 15: if visible is false, is there a way for the user to make it visible?

should we also mark the API keys as sensitive so that they don't show on screen or is that overkill?

sva91: I have updated the variables to do the following:
1. By default, they will be hidden. 
2. If the advanced setting "Allow Datadog Environment Variable Changes" is selected, variables will become visible, and are now modifiable.

datadog-integration/schema.yaml
```diff
@@ -12,31 +12,27 @@ variableGroups:
...
+    visible: ${allow_datadog_env_changes}
 
 variables:
-# Tenancy
-  tenancy_ocid:
-      visible: false
-
-  region:
-    visible: false
+  # Advanced Options
```
maxpyf, line 27: This is great. I would just change the wording a little.

title: Override Datadog Variables
description: Click to view/modify pre-set Stack variables

maxpyf (approved):
JUst one wording nit otherwise ship it!
And thank you!

datadog-integration/schema.yaml
sva91: Done

PR #61, opened 2025-05-28
add pre checks

sva91:
## What

- Adds 5 validation checks
- Terraform plan and apply jobs will fail if validations fail
- If the validations succeed once, we don't recheck them again, unless a destroy job is executed
- Validation error will be printed in logs

## Why
- Fail fast if some common issues occur

## Testing


datadog-integration/pre_check.py
```diff
@@ -0,0 +1,213 @@
+from enum import Enum
+import sys
+import json
+import subprocess
+import os
+
+DEFAULT_DOMAIN_NAME = "Default"
+MIN_VAULT_QUOTA = 1
+MIN_CONNECTOR_HUB_QUOTA_PER_REGION = 1
```
maxpyf, line 9: is 1 really enough?

maxpyf (approved):
nice work!
just take a look at what the bot flagged.

datadog-integration/pre_check.py
```diff
@@ -0,0 +1,213 @@
...
+    supported_regions = json.loads(params.get("supported_regions", "[]"))
+
+    errors = []
+    # Find domain name and domain endpoint for further checks
+    domain_name, domain_endpoint = find_user_domain(user_id, tenancy_ocid)
+
+    # Validation 1: Home region check
+    is_home_region = params.get("is_home_region", "false").lower() == "true"
+    result = validate_home_region(is_home_region)
+    if result != "ok":
```
kanishktripathi, line 180: nit: `"ok"` can be a constant

datadog-integration/pre_check.py
sva91: Not sure, I added the code to make some validations, and kept it as 1 for now. We can discuss what the right number should be, which is applicable to all tenants.

PR #67, opened 2025-06-06
ECI-684: add support for identity domains and advanced inputs

maxpyf:
## What

ECI-650, ECI-684, ECI-649: changes to allow "advanced" parameters as inputs.
This is meant to be run from OCI RM Stack.
1. Non default domain usage: the domain is the currently logged in domain. Default is no longer assumed.
2. Choice of subnet for forwarding functions: allow input of a list of subnet OCIDs, one per region. The final set of regional stacks is an intersection of (tenancy subscribed regions, domain replicated regions, input subnet regions]. This input is optional. If left empty (default), the stack creates a VCN with subnets in all subscribed regions.
3. Choice of Compartment for deploying stack resources. If left empty, we will create/use the "Datadog" compartment.
4. Choice of Group/User. If left empty (default), we create a User and Group. Otherwise, we use the supplied User & Group. A sizeable amount of the code in this PR revolves around validating these inputs (ensure User belongs to Group, either both are null or neither is, ensure Group belongs to domain).

## Why

Multiple requests.

## How was it tested?

Manually.

### DIFF
--- a/datadog-integration/modules/auth/locals.tf
+++ b/datadog-integration/modules/auth/locals.tf
@@ -1,0 +1,3 @@
+locals {
+  email = var.user_email
+}
--- a/datadog-integration/modules/auth/main.tf
+++ b/datadog-integration/modules/auth/main.tf
@@ -3,76 +3,211 @@ terraform {
   required_providers {
     oci = {
       source  = "oracle/oci"
-      version = "5.46.0"
+      version = ">=7.1.0"
     }
   }
 }
 
-resource "oci_identity_user" "dd_auth" {
-  # Required
-  compartment_id = var.tenancy_id
-  description    = "[DO NOT REMOVE] Read only user created for fetching resources metadata which is used by Datadog Integrations"
-  name           = var.user_name
-  email          = var.email
-  freeform_tags  = var.tags
+# Stage 1: Validate that both user and group variables are either both null/empty OR both provided
+resource "null_resource" "user_group_variable_validation" {
+  
+  provisioner "local-exec" {
+    when       = create
+    on_failure = fail
+    command = <<-EOT
+      # Check variable consistency
+      USER_PROVIDED="${var.existing_user_id != null && var.existing_user_id != "" ? "true" : "false"}"
+      GROUP_PROVIDED="${var.existing_group_id != null && var.existing_group_id != "" ? "true" : "false"}"
+      
+      if [ "$USER_PROVIDED" != "$GROUP_PROVIDED" ]; then
+        echo "ERROR: Both existing_user_id and existing_group_id must be provided together, or both must be null/empty."
+        echo "Current state:"
+        echo "  existing_user_id: ${var.existing_user_id != null ? var.existing_user_id : "null/empty"}"
+        echo "  existing_group_id: ${var.existing_group_id != null ? var.existing_group_id : "null/empty"}"
+        exit 1
+      fi
+      
+      if [ "$USER_PROVIDED" = "false" ] && [ "$GROUP_PROVIDED" = "false" ]; then
+        echo "✅ No existing user/group provided - will create new ones"
+      else
+        echo "✅ Both existing user and group provided - will validate them in next stage"
+      fi
+    EOT
+  }
 }
 
-resource "oci_identity_group" "dd_auth" {
-  # Required
-  compartment_id = var.tenancy_id
-  description    = "[DO NOT REMOVE] Group for adding permissions to Datadog user"
-  name           = var.user_group_name
-  freeform_tags  = var.tags
+# Stage 2: Validate data sources only when both variables are provided
+resource "null_resource" "existing_user_group_validation" {
+  count = var.existing_user_id != null && var.existing_user_id != "" && var.existing_group_id != null && var.existing_group_id != "" ? 1 : 0
+  
+  depends_on = [null_resource.user_group_variable_validation]
+  
+  provisioner "local-exec" {
+    when       = create
+    on_failure = fail
+    command = <<-EOT
+      # First check if the data sources returned any results
+      USER_DATA_COUNT=${length(data.oci_identity_domains_users.existing_user_with_groups)}
+      GROUP_DATA_COUNT=${length(data.oci_identity_domains_groups.existing_group)}
+      
+      if [ $USER_DATA_COUNT -eq 0 ]; then
+        echo "ERROR: No user data found for OCID '${var.existing_user_id}'"
+        echo "This could mean:"
+        echo "  - The OCID is invalid or malformed"
+        echo "  - The user does not exist in this Identity Domain"
+        echo "  - The OCID belongs to a different Identity Domain"
+        echo "  - You don't have permission to read this user"
+        exit 1
+      fi
+      
+      if [ $GROUP_DATA_COUNT -eq 0 ]; then
+        echo "ERROR: No group data found for OCID '${var.existing_group_id}'"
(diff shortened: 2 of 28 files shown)


maxpyf:
* **#67** <a href="https://app.graphite.dev/github/pr/DataDog/oracle-cloud-integration/67?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/oracle-cloud-integration/67?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

datadog-integration/modules/regional-stacks/locals.tf
```diff
@@ -21,4 +21,43 @@ locals {
   nat_gateway     = "${local.vcn_name}-natgateway"
   service_gateway = "${local.vcn_name}-servicegateway"
   subnet          = "${local.vcn_name}-private-subnet"
+
```
maxpyf, line 24: allows searching for a subnet by partial name, anywhere in the tenancy.

datadog-integration/modules/compartment/main.tf
```diff
@@ -3,14 +3,41 @@ terraform {
...
+
+  lifecycle {
+    # Prevent destruction of the compartment
+    prevent_destroy = true
+    # Ignore changes to avoid conflicts
+    ignore_changes = [
+      description,
+      freeform_tags
+    ]
+  }
```
sva91, line 26: Compartments are not destroyed by terraform. Do we need this for any other reason?

datadog-integration/modules/compartment/main.tf
```diff
@@ -3,14 +3,41 @@ terraform {
...
+    name   = "name"
+    values = [var.compartment_name]
+  }
+  
+  depends_on = [oci_identity_compartment.this]
+}
+
+# Local to output the compartment ID
+locals {
+  compartment_id = data.oci_identity_compartments.this.compartments[0].id
```
sva91, line 42: Don't think we need this

datadog-integration/modules/compartment/outputs.tf
```diff
@@ -1,3 +1,3 @@
 output "id" {
-  value = oci_identity_compartment.this.id
+  value = local.compartment_id
```
sva91, line 2: we can probably output directly from resource

datadog-integration/modules/domain/main.tf
```diff
@@ -0,0 +1,7 @@
+data "oci_identity_domains" "domains" {
+  compartment_id = var.tenancy_id
+}
+
+data "oci_identity_domain" "domain" {
+  domain_id = [for domain in data.oci_identity_domains.domains.domains : domain.id if domain.display_name == var.domain_name][0]
+}
```
sva91, line 7: Do we need a new module here if we aren't creating any resources ? We can probably just do this in root module.

sva91: We can reuse this data block:  https://github.com/DataDog/oracle-cloud-integration/blob/1e0a18899df2a590a206b81eda9e6c91b6060750/datadog-integration/data.tf#L15

datadog-integration/modules/regional-stacks/main.tf
```diff
@@ -3,20 +3,32 @@ terraform {
   required_providers {
     oci = {
       source  = "oracle/oci"
-      version = "5.46.0"
+      version = ">=7.1.0, <7.4.0"
```
sva91, line 6: curious why put limitation on the max version ? Is there anything broken above 7.4.0 ?

datadog-integration/modules/auth/locals.tf
```diff
@@ -0,0 +1,8 @@
+locals {
+  user_group_name  = "dd-svc-user-group"
+  user_policy_name = "dd-svc-user-policy"
+  dg_sch_name      = "dd-dynamic-group-connectorhubs"
+  dg_fn_name       = "dd-dynamic-group-functions"
+  dg_policy_name   = "dd-dynamic-group-policy"
+  email            = data.oci_identity_domains_user.current_user.emails[0].value
```
sva91, line 7: these are duplicated from the root module https://github.com/DataDog/oracle-cloud-integration/blob/1e0a18899df2a590a206b81eda9e6c91b6060750/datadog-integration/locals.tf#L19

It would be best if we can specify them in one location, and pass it to the sub-module, to avoid any mistakes later.

datadog-integration/modules/auth/main.tf
```diff
@@ -3,76 +3,83 @@ terraform {
...
 }
 
-resource "oci_identity_user" "dd_auth" {
-  # Required
-  compartment_id = var.tenancy_id
-  description    = "[DO NOT REMOVE] Read only user created for fetching resources metadata which is used by Datadog Integrations"
-  name           = var.user_name
-  email          = var.email
-  freeform_tags  = var.tags
+resource "oci_identity_domains_user" "dd_auth" {
```
sva91, line 11: Can't we specify tags for this resource ?

datadog-integration/modules/auth/main.tf
```diff
@@ -3,76 +3,83 @@ terraform {
...
 }
 
-resource "oci_identity_group" "dd_auth" {
-  # Required
-  compartment_id = var.tenancy_id
-  description    = "[DO NOT REMOVE] Group for adding permissions to Datadog user"
-  name           = var.user_group_name
-  freeform_tags  = var.tags
-}
+resource "oci_identity_domains_group" "dd_auth" {
```
sva91, line 28: Same, can we specify tags ?

sva91:
This looks pretty good so far, and I am very happy that this works !! We probably should also bump config_version [here](https://github.com/DataDog/oracle-cloud-integration/blob/1e0a18899df2a590a206b81eda9e6c91b6060750/datadog-integration/modules/integration/locals.tf#L3), if we do get some customers on version 2.

datadog-integration/modules/compartment/main.tf
maxpyf: cursor addition that sneaked by...

datadog-integration/modules/regional-stacks/main.tf
maxpyf: I had a runtime issue on oci with a pre-existing stack that suddenly started using 7.4.0. The limitation will be removed.

datadog-integration/modules/auth/locals.tf
maxpyf: my guess is they are not needed in the root module, since it doesn't do anything with them.
I just had an issue with `${local.user_name}` which expanded to an invalid name in my case.

datadog-integration/modules/auth/main.tf
maxpyf: we can, but it took some doing because this resource uses the scim protocol.

datadog-integration/modules/auth/main.tf
maxpyf: same as above

maxpyf: I think it works now. these providers use a very different way to pass tags so it was a bit confusing.

bengillmandd (approved):

PR #72, opened 2025-07-08
bug fix: ensure final region set only includes regions where docker image is available

maxpyf:
The final region set was the intersection of user input, domain supported and tenancy supported regions.
We forgot the check for availability of docker images in all the recent changes.

### DIFF
--- a/datadog-integration/locals.tf
+++ b/datadog-integration/locals.tf
@@ -134,10 +134,19 @@ locals {
     ]) > 1
   ]
 
-  # Intersection of subscribed regions, regions in domain, and subnet regions (compatible with Terraform <= 1.5.x)
+  docker_image_enabled_regions = toset([
+    for region in keys(local.subscribed_regions_list) : region
+    if local.supported_regions[region].result.failure == ""
+  ])
+
+  # Intersection of subscribed regions, regions in domain, subnet regions (compatible with Terraform <= 1.5.x)
+  # and regions where the docker image is available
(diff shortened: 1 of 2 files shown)


kanishktripathi (approved):

kanishktripathi (approved):

datadog-integration/locals.tf
```diff
@@ -134,10 +134,24 @@ locals {
...
+
+  # Intersection of subscribed regions, regions in domain, subnet regions (compatible with Terraform <= 1.5.x)
   # Only create regional stacks for regions that meet all three criteria
   target_regions_for_stacks = length(local.subnet_ocids_list) > 0 ? toset([
     for region in local.subscribed_regions_list : region
     if contains(tolist(local.regions_in_domain_set), region) && contains(tolist(local.subnet_regions), region)
-  ]) : local.subscribed_regions_set
+  ]) : toset([
+    for region in local.subscribed_regions_list : region if contains(tolist(local.regions_in_domain_set), region)
+  ])
```
sva91, line 149: Not fully understanding why this else block had to change. Is it for the scenario where advanced users provide a different domain, but no subnets ?

maxpyf: yes exactly. I think it was a bug in that clause, we would just use all the regions in the tenancy. 
I am not sure if it is really needed to be honest, but if the domain has missing regions, I don't think the user will be able to see them, so it will limit our visibility.

PR #73, opened 2025-07-11
Update docker image creation script to add for specific region

kanishktripathi:
**what:**
* Add option to push docker image for a specific region.

**why:**
* By default the image is added to all region. In case of a newly provisioned region, it will be good to use this option to only add image for that specific region


sva91 (approved):
LGTM

PR #111, opened 2026-04-28
ECI-1621: Add events-forwarder OCI Function

yuhuyoyo:
## What

- **`client.go`**: Added variadic `extraHeaders ...map[string]string` to `SendMessageToDatadog` and `sendMessage`. Callers can inject arbitrary HTTP headers without breaking the logs/metrics forwarders (backward-compatible, no-arg path unchanged).
- **`handler.go`**: Removes the `formatter.Stamp()` call entirely — OCI CloudEvents are forwarded raw with no per-event JSON marshal/unmarshal. Injects a `Dd-Oci-Tenancy-Id` HTTP header carrying the tenancy OCID on every POST to intake.
- **`formatter.go`**: Removes the `Stamp()` function and its file-level doc; package comment updated to reflect events are forwarded unchanged.
- **`formatter_test.go`**: Removes the 4 now-obsolete Stamp tests.
- **`handler_test.go`**: Splits the missing-config test into `TestMyHandler_MissingTenancyOCID` and `TestMyHandler_MissingDDSite`; sets `TENANCY_OCID` env var in `withTestClient`.

## Why

**No per-event marshal/unmarshal**: OCI Streaming → Connector Hub is a consumer group that can pull up to 4 MB (~4000 events) per invocation. Stamping each event with `json.Unmarshal` + field injection + `json.Marshal` in a tight loop would be costly for bursty traffic. Forwarding raw bytes sidesteps this entirely.

**Tenancy OCID via header**: Rather than embedding the tenancy OCID in each event's JSON body, we inject it as `Dd-Oci-Tenancy-Id` on the HTTP request. The intake preserves request headers in the `evp.intake` Kafka header, making them available to the cloudchanges worker downstream without touching the payload.

**Not overwriting the native `source` field**: OCI CloudEvents carry a meaningful `source` field (e.g. `"objectstorage"`, an OCID). Stamping `source:"oci"` would clobber that. The dd-source worker (DataDog/dd-source#435907) now uses `cloudEventsVersion` as the sole OCI routing signal, so no source stamp is needed.

## Test plan

- [ ] `go test ./...` passes in `datadog-functions/events-forwarder/`
- [ ] `Dd-Oci-Tenancy-Id` header appears on intake requests in staging
- [ ] dd-source counterpart PR: https://github.com/DataDog/dd-source/pull/435907


datadog-functions/events-forwarder/internal/formatter/formatter.go
```diff
@@ -0,0 +1,109 @@
...
+// inspection. Customer events forwarded directly to the v2 API (no stamp) fall
+// back to cloudEventsVersion detection on the worker side.
+func Stamp(events []json.RawMessage) ([]json.RawMessage, error) {
+	stamped := make([]json.RawMessage, 0, len(events))
+	for _, ev := range events {
+		var m map[string]json.RawMessage
+		if err := json.Unmarshal(ev, &m); err != nil {
+			return nil, fmt.Errorf("failed to parse event for stamping: %w", err)
+		}
+		m["source"] = json.RawMessage(`"oci"`)
```
EamonBrady1: I assume there's no "source" field already that we would be overwriting?

datadog-functions/events-forwarder/internal/formatter/formatter.go
```diff
@@ -0,0 +1,109 @@
...
+		}
+		buf, _ := json.Marshal(current)
+		payloads = append(payloads, buf)
+		current = current[:0]
+		currentSize = 2
+	}
+
+	for _, ev := range events {
+		evSize := len(ev)
+		if evSize+2 > MaxBodyBytes {
```
EamonBrady1, line 66: Do we just drop events that are larger than our maximum allowed size? Does this ever happen (I assume not) and anything we can do to split up a message of that size?

datadog-functions/events-forwarder/internal/handler/handler.go
```diff
@@ -0,0 +1,52 @@
...
+
+func MyHandler(ctx context.Context, in io.Reader, out io.Writer) {
+	events, err := formatter.Decode(in)
+	if err != nil {
+		log.Printf("Error decoding event batch: %v", err)
+		writeResponse(out, "error", "", err)
+		return
+	}
+
+	ddclient, site, err := datadogClientFunc()
```
EamonBrady1: It looks like this returns tenancy OCID too?:
https://github.com/DataDog/oracle-cloud-integration/blob/master/datadog-functions/metrics-forwarder/internal/handler/handler.go#L38

Do events already include that? otherwise might be good to enrich if we're already unmarshaling

datadog-functions/events-forwarder/internal/formatter/formatter.go
EamonBrady1: I see we do drop it. That seems reasonable since there would be no good way to send a half event, and it's probably a safe assumption we won't run into this

EamonBrady1 (commented):
Any concerns about unmarshaling and re-marshaling bursty traffic?

datadog-functions/events-forwarder/internal/formatter/formatter.go
yuhuyoyo: good catch. there is a source field. I'm renaming this to ddForwarder and also adding the tenancy ocid.

datadog-functions/events-forwarder/internal/formatter/formatter.go
yuhuyoyo: by avg an event is about 1kb.

EamonBrady1 (approved):
LGTM

PR #114, opened 2026-04-30
fix: vault-quota precheck blocks idempotent re-apply (issue #110)

yuhuyoyo:
## What

- **`datadog-terraform-onboarding/prechecks.tf`**: Fix `validate_vault_quota` precondition to short-circuit when the vault already exists in state, making re-applies idempotent. Introduces `check_resources_in_state` (replaces unused `check_integration_exists`) which runs `terraform state list` once and detects quota-guarded resources by type pattern rather than hardcoded module address.
- **`datadog-terraform-onboarding/prechecks.tf`**: Remove `validate_connector_hub_quota` — this module creates no `oci_sch_service_connector` resource; the connector lives in `datadog-oci-orm/metrics-setup` which has its own state. The check guarded a resource it didn't manage and could never detect as already-existing.

## Why

Fixes #110. The `validate_vault_quota` precondition (`available >= 1`) fired on every `plan`/`apply`. After first apply the dedicated Datadog vault consumed the per-region `key-management/virtual-vault-count` quota. In tenancies where that quota is 1 (the default), `available` dropped to 0 and every subsequent plan aborted with the vault-quota exhausted error — even though the vault already existed in state and no new vault would be created.

The fix makes the check idempotent: quota headroom **or** vault already in state → precondition passes. The error still fires correctly on a fresh deploy when quota is genuinely exhausted.

## Test plan

- [ ] `terraform validate` in `datadog-terraform-onboarding/` passes
- [ ] Fresh deploy with quota=0 still fails with the vault-quota error
- [ ] Re-apply in a tenancy at quota limit passes (vault found in state)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #116, opened 2026-05-04
ECI-1626: pre_check.py quota checks block idempotent re-apply

yuhuyoyo:
Mirrors the fix from PR #114 (datadog-terraform-onboarding/prechecks.tf) for the datadog-integration / ORM stack flow.

## Changes

- **validate_vault_quota**: short-circuit when `datadog-vault` already exists in OCI (looked up via `oci kms management vault list` in the resolved compartment). On a re-apply where the vault was created by a prior apply the per-region virtual-vault-count quota is already consumed, so the raw quota check would incorrectly abort. The fix passes when quota >= 1 **or** the vault already exists.

- **validate_connector_hub_quota**: removed. The service connector is created by the regional OCI Resource Manager nested stacks (datadog-oci-orm/metrics-setup), not by this stack — the check guarded a resource it doesn't manage, exactly as #114 noted for the Terraform-onboarding path.

- **validate_pre_existing_resources** (policy checks): on re-apply, `dd-svc-policy` and `dd-dynamic-group-policy` already exist in OCI because Terraform created them on the first apply. The pre-check was erroring unconditionally on their existence. Fix adds `_policy_owned_by_datadog()` which inspects the `freeform-tags.ownedby` tag that the auth module stamps on every IAM policy it creates. A policy is only flagged as a collision if it exists **and** lacks that tag — i.e. it wasn't created by this stack.

- **main.tf**: pass `--compartment-id` to pre_check.py so the vault lookup works both when the user supplies an existing compartment and when the stack auto-creates one named "Datadog".

## Out of scope

The IDCS resource checks (user, group, dynamic groups) in `validate_pre_existing_resources` have the same re-apply false positive risk in theory, but are unaffected in practice for most tenancies because the dynamic group resources currently carry no ownership tags in the auth module. That fix is deferred to a follow-up.

### DIFF
--- a/datadog-integration/pre_check.py
+++ b/datadog-integration/pre_check.py
@@ -8,7 +8,8 @@ OK_STATUS = "ok"
 ERROR_STATUS = "error"
 DEFAULT_DOMAIN_NAME = "Default"
 MIN_AVAILABLE_VAULT = 1
-MIN_AVAILABLE_CONNECTOR_HUB = 1
+DATADOG_VAULT_NAME = "datadog-vault"
+DATADOG_COMPARTMENT_NAME = "Datadog"
 
 class ResourceType(Enum):
     USER = "user"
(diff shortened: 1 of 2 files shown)


datadog-integration/pre_check.py
```diff
@@ -8,7 +8,8 @@
 ERROR_STATUS = "error"
 DEFAULT_DOMAIN_NAME = "Default"
 MIN_AVAILABLE_VAULT = 1
-MIN_AVAILABLE_CONNECTOR_HUB = 1
```
sva91, line 11: Any reason to remove this check ? I think we put it in there because we wanted to be able to create at-least one connector hub always. Without being able to do that, the integration won't collect any data, and will appear broken, even though it gets created successfully.

datadog-integration/pre_check.py
```diff
@@ -71,19 +72,71 @@ def validate_pre_existing_resources(params, domain_endpoint):
...
-    if _resource_exists(ResourceType.POLICY, params["user_group_policy_name"], compartment_id=params["tenancy_id"]):
+    if _resource_exists(ResourceType.POLICY, params["user_group_policy_name"], compartment_id=params["tenancy_id"]) \
+            and not _policy_owned_by_datadog(params["user_group_policy_name"], params["tenancy_id"]):
         existing.append(f"User Group Policy {params['user_group_policy_name']}")
     if _resource_exists(ResourceType.DYNAMIC_GROUP, params["dg_sch_name"], domain_endpoint=domain_endpoint):
         existing.append(f"Dynamic Group {params['dg_sch_name']}")
     if _resource_exists(ResourceType.DYNAMIC_GROUP, params["dg_fn_name"], domain_endpoint=domain_endpoint):
         existing.append(f"Dynamic Group {params['dg_fn_name']}")
-    if _resource_exists(ResourceType.POLICY, params["dg_policy_name"], compartment_id=params["tenancy_id"]):
+    if _resource_exists(ResourceType.POLICY, params["dg_policy_name"], compartment_id=params["tenancy_id"]) \
```
sva91, line 82: that's a good addition. thanks. do we need this in terraform setup too ?

datadog-integration/pre_check.py
yuhuyoyo: my understanding is that connectorhubs are created by hubmanager and not the stack. So I think it makes more sense to just rely on Integration health service to check for available connectorhub quota than here as we don't manage connectorhub states via terraform.

Also on reapply, if the connectorhubs were already created by hubmanager but the customer's quota at the given region is at the limit, we will have false positive here in the check.

datadog-integration/pre_check.py
yuhuyoyo: the terraform set up only check for terraform state and not the actual resources. it doesn't have the equivalent of _resource_exists either.

sva91 (approved):

PR #118, opened 2026-05-05
feat: release workflow using dd-octo-sts

yuhuyoyo:
## Summary

- Adds `.github/workflows/release.yml` — a `workflow_dispatch` workflow that automates releases by pushing tags and creating GitHub releases via dd-octo-sts
- Replaces manual tag creation, which is blocked by tag protection rules
- Accepts a `bump` input (patch/minor/major) and auto-computes the next version from the latest `datadog-oci-orm-vX.Y.Z` tag
- Tag and release are created via the GitHub API using a short-lived dd-octo-sts token (no checkout needed)

## Why

Tag protection rules block manual tag creation. The workflow exchanges a short-lived OIDC token for a GitHub App token with `contents: write` scoped to this repo via dd-octo-sts, bypassing the protection in a controlled, auditable way.

## Merge order

**Requires the trust policy PR (#117) to be merged to `master` first.** dd-octo-sts reads the trust policy from the default branch at token exchange time.

## Known limitation

If tag creation succeeds but `gh release create` fails, retrying will compute the next version from the newly-created orphan tag, skipping a version number. Delete the orphan tag before retrying in that case.

## Test plan

- [x] Merge PR #117 (trust policy) to `master`
- [ ] Merge this PR to `master`
- [ ] Go to Actions → Release → Run workflow, select `patch`
- [ ] Confirm new tag `datadog-oci-orm-v1.1.7` and release are created

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,10 +3,14 @@ name: Release
 on:
   workflow_dispatch:
     inputs:
-      version:
-        description: "Version tag to create (e.g. datadog-oci-orm-v1.6.0)"
+      bump:
+        description: "Version component to bump"
         required: true
-        type: string
+        type: choice
+        options:
+          - patch
+          - minor
+          - major
 
 jobs:
   release:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,104 @@
...
+            --raw-field "ref=refs/tags/${VERSION}" \
+            --raw-field "sha=${GITHUB_SHA}"
+
+      - name: Create release
+        env:
+          VERSION: ${{ steps.version.outputs.version }}
+          GH_TOKEN: ${{ steps.octo-sts.outputs.token }}
+          GH_REPO: ${{ github.repository }}
+        run: |
+          gh release create "${VERSION}" \
```
sva91: I think this marks the release as "latest" immediately with no assets attached. Customers rely on the releases/latest/download/<file> URL pattern to fetch installation zips, they'll hit 404s until assets are manually uploaded. Consider either building and uploading assets in this same workflow before this step, or passing --latest=false and promoting to latest only after assets are confirmed attached.

sva91 (changes requested):

sva91 (approved):
LGTM

Issue #109, opened 2026-04-28
`dd-metrics-forwarder` and `dd-logs-forwarder` error with `FunctionInvokeImageNotAvailable`


florianmutter:
### Summary                                               

  In `datadog-oci-orm-v1.1.6` (and master), the `regional-stacks` submodule has two Terraform resources that both call `manage_default_resource_id` on the same VCN default security list:                                                                                                                                
   
  1. `module.vcn[0].oci_core_default_security_list.lockdown` (inside `oracle-terraform-modules/vcn/oci`, enabled because the Datadog module passes `lockdown_default_seclist = true`) — writes **no** egress/ingress rules and has `ignore_changes = [egress_security_rules, ingress_security_rules, defined_tags]`.      
  2. `oci_core_default_security_list.dd_default` (in `modules/regional-stacks/main.tf`) — writes egress-all + two ICMP ingress rules.
                                                                                                                                                                                                                                                                                                                          
  On `terraform apply`, both fire `UpdateSecurityList` PUTs back-to-back. In our deployment the `lockdown` resource won the race, leaving the default SL with empty `egress_security_rules`. The forwarder Functions then could not pull their images from OCIR, and every Service Connector Hub invocation returned:     
                                                                                                                                                                                                                                                                                                                          
  ```                                                                                                                                                                                                                                                                                                                     
  {"code":"FunctionInvokeImageNotAvailable","message":"Failed to pull function image"}  (502)
  ```                                                                                                                                                                                                                                                                                                                     
   
  This affects both `dd-metrics-forwarder` and `dd-logs-forwarder`.                                                                                                                                                                                                                                                       
                                                            
  ### How we found it                                                                                                                                                                                                                                                                                                     
                                                            
  OCI Monitoring (`oci_faas` namespace) showed `FunctionResponseCount` with dimensions:                                                                                                                                                                                                                                   
                                                            
  ```                                                                                                                                                                                                                                                                                                                     
  ErrorCode = 502                                           
  ErrorMessage = FunctionInvokeImageNotAvailable
  ```                                                                                                                                                                                                                                                                                                                     
   
  for both functions and only that error class.                                                                                                                                                                                                                                                                           
                                                            
  The OCIR images themselves were healthy (anonymous token + manifest fetch returned 200 with the exact digests pinned on the Functions). The route table correctly sent `all-fra-services-in-oracle-services-network` to the service gateway. The only thing missing was the egress rule on the SL.                      
                                                            
[...]

PR #119, opened 2026-05-05
ECI-1625: (fix) add depends_on to prevent race condition in default security list

yuhuyoyo:
## What

Add `depends_on = [module.vcn]` to `oci_core_default_security_list.dd_default` in both `datadog-integration` and `datadog-terraform-onboarding` regional-stacks modules.

## Why

PR #99 (CLOUDS-7599) introduced a race condition where two Terraform resources simultaneously manage the same VCN default security list via `manage_default_resource_id`:

- **Resource 1**: `module.vcn[0]` (upstream `oracle-terraform-modules/vcn/oci`) internally creates `oci_core_default_security_list.lockdown`, which writes **empty rules** (no egress, no ingress) and has `lifecycle { ignore_changes = [...] }`.
- **Resource 2**: `oci_core_default_security_list.dd_default` — writes the correct minimal rules: egress-all (required for OCIR image pulls) + two ICMP-only ingress rules.

Both target the same security list ID. The OCI audit log shows two `UpdateSecurityList` calls ~158ms apart during apply. If `lockdown` wins the race, the list is left with empty egress — functions cannot reach OCIR → 502 `FunctionInvokeImageNotAvailable`. Because `lockdown` has `ignore_changes`, Terraform won't self-correct on subsequent applies.

`depends_on = [module.vcn]` enforces ordering: the entire VCN module (including its hidden `lockdown` resource) completes before `dd_default` writes the correct rules. Since `lockdown` has `ignore_changes`, it won't overwrite `dd_default` afterward.

## References

- Jira: [ECI-1625](https://datadoghq.atlassian.net/browse/ECI-1625)
- GitHub issue: #109
- Introducing PR: #99

### DIFF
--- a/datadog-integration/modules/regional-stacks/main.tf
+++ b/datadog-integration/modules/regional-stacks/main.tf
@@ -75,6 +75,7 @@ module "subnet" {
 resource "oci_core_default_security_list" "dd_default" {
   count                      = var.subnet_ocid == "" ? 1 : 0
   manage_default_resource_id = data.oci_core_vcn.dd_vcn[0].default_security_list_id
+  depends_on                 = [module.vcn]
   freeform_tags              = var.tags
 
   egress_security_rules {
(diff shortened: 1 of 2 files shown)


sva91 (approved):
LGTM

PR #121, opened 2026-05-05
Fix dry-run release workflow: remove --cleanup-tag from draft delete

yuhuyoyo:
## Summary

- Remove `--cleanup-tag` from the dry-run delete step: GitHub does not create a tag for draft releases, so the flag was failing with `HTTP 422: Reference does not exist`
- Switch version prefix from `datadog-oci-orm-v` to `datadog-integration-v`: the `datadog-oci-orm-v` series had a stray `v1.5.5` tag that was causing the next computed version to jump to `v1.5.6` instead of continuing from the intended `v1.1.x` series; the `datadog-integration-v` tags top out at `v1.1.4` so the next release will be `v1.1.5`

## Root cause (dry-run failure)

Run [#25396494703](https://github.com/DataDog/oracle-cloud-integration/actions/runs/25396494703/job/74484581467) failed at "Delete draft release and tag" with:
```
HTTP 422: Reference does not exist (https://api.github.com/repos/DataDog/oracle-cloud-integration/git/refs/tags/datadog-oci-orm-v1.5.6)
```

The draft release was created successfully and deleted, but `--cleanup-tag` blew up since no tag ever existed (draft releases don't create tags, and "Create tag" is intentionally skipped in dry-run mode).

## Test plan

- [ ] Run the release workflow with `dry_run: true` and confirm it completes successfully with version `datadog-integration-v1.1.5`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,7 +12,7 @@ on:
           - minor
           - major
       dry_run:
-        description: "Dry run: build and upload assets to a draft release, then delete it. An ephemeral tag is created and deleted."
+        description: "Dry run: build and upload assets to a draft release, then delete it."
         type: boolean
         default: false
 
@@ -129,14 +129,14 @@ jobs:
         run: |
           gh release edit "${VERSION}" --latest --repo "${GH_REPO}"
 
-      - name: Delete draft release and tag
+      - name: Delete draft release
         if: always() && inputs.dry_run == true && steps.create-release.conclusion == 'success'
         env:
           VERSION: ${{ steps.version.outputs.version }}
           GH_TOKEN: ${{ steps.octo-sts.outputs.token }}
           GH_REPO: ${{ github.repository }}
         run: |
-          gh release delete "${VERSION}" --repo "${GH_REPO}" --yes --cleanup-tag
+          gh release delete "${VERSION}" --repo "${GH_REPO}" --yes
 
       - name: Delete release and tag on failure
         if: steps.create-tag.conclusion == 'success' && failure()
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -41,16 +41,16 @@ jobs:
           fi
 
           latest=$(gh api "repos/${GH_REPO}/tags" --paginate --jq '.[].name' \
-            | grep -E '^datadog-oci-orm-v[0-9]+\.[0-9]+\.[0-9]+$' \
+            | grep -E '^datadog-integration-v[0-9]+\.[0-9]+\.[0-9]+$' \
```
sva91, line 44: i am not sure I understand this change. From the release, the name is datadog-oci-orm-vxxx - https://github.com/DataDog/oracle-cloud-integration/releases 
I am not sure if we can change the name without breaking redirect link from the tile.

yuhuyoyo: the naming is not very consistent https://github.com/DataDog/oracle-cloud-integration/releases/tag/datadog-integration-v1.1.4 this one is datadog-integration-v1.1.4. and then we have a https://github.com/DataDog/oracle-cloud-integration/tree/**datadog-oci-orm-v1.5.5** which is before https://github.com/DataDog/oracle-cloud-integration/tree/**datadog-oci-orm-v1.1.6**. i'm not aware we have a link in the tile. how is that defined?

sva91: I see some code reference here : https://github.com/DataDog/web-ui/blob/899fcba48d7bad9def8f0175ed7d08b0464fdfc6/packages/apps/integrations/toolkit/eci/oracle-cloud-infrastructure.constants.ts#L43

sva91: looks like we just need `datadog-integration.zip` to be present in the latest release for the redirect to work. the release name doesn't matter much here apparantly, which is good probably.

sva91 (approved):

PR #125, opened 2026-05-12
[ECI-1615] Add IAM for HubManager to manage the events forwarding pipeline

yuhuyoyo:
## What

Adds the IAM primitives the Datadog group (`dd-svc`) needs to stand up and tear down the OCI events forwarding pipeline out-of-band when a customer opts in / out via the Datadog UI. Applies the same set of changes to both stacks (`datadog-integration/` and `datadog-terraform-onboarding/`).

The pipeline resources themselves (function, stream, events rule, connector hub) are **not** created by this stack — Datadog creates them at runtime. This PR only adds the IAM so that flow works.

## Why

ECI-1615 / M1 of the OCI events RFC. The events flow requires:

- The Datadog group (acting as `dd-svc`) to create + delete the events rule, events stream, and forwarder function on opt-in / opt-out
- The SCH dynamic group to pull from the events stream and invoke the function
- OCI Events Service (acting as the `eventrule` service principal) to publish to the events stream when a rule fires

…all without granting the Datadog group the ability to touch resources it doesn't own.

## How resources are scoped

A `DatadogManaged.marker` defined tag (namespace + tag created in this PR) acts as the marker for Datadog-owned resources. Datadog tags every rule / stream it creates. Each grant uses a `where` clause that pairs **CREATE-via-`request.permission`** with **UPDATE/DELETE/READ-via-`target.resource.tag`** — this is the only way to enforce the tag because OCI's `target.resource.tag` doesn't evaluate against the request body on CREATE (the resource doesn't exist yet to read tags from).

### New resources

- `oci_identity_tag_namespace.datadog_managed` — `DatadogManaged` namespace in the Datadog compartment
- `oci_identity_tag.marker` — `marker` tag inside that namespace

### `oci_identity_policy.dd_auth` (Datadog group)

- `manage cloudevents-rules in tenancy where any { request.permission = 'EVENTRULE_CREATE', target.resource.tag.DatadogManaged.marker = 'true' }` — Datadog creates / deletes the events rule
- `manage streams in compartment where any { request.permission = 'STREAM_CREATE', target.resource.tag.DatadogManaged.marker = 'true' }` — Datadog creates / deletes the events stream (OCI streaming has no `target.stream.name` variable, so tag-based scoping is the only narrow option)

### `oci_identity_policy.dynamic_group` (runtime data path)

- `service_connector use stream-pull in compartment where target.resource.tag.DatadogManaged.marker = 'true'` — SCH consumes from the events stream
- `any-user use stream-push in compartment where all { request.principal.type = 'eventrule', target.resource.tag.DatadogManaged.marker = 'true' }` — OCI Events Service publishes to the events stream. `any-user` is required because `eventrule` is a service principal (no named group / dynamic-group represents it); the principal-type and target-tag conditions together restrict this to events-service publishes against Datadog-managed streams.

## Trade-off: CREATE is broader than UPDATE/DELETE

For both `cloudevents-rules` and `streams`, the `request.permission = '<X>_CREATE'` branch allows CREATE on *any* rule/stream in scope, not just ones being tagged. We can't enforce the tag on CREATE because OCI doesn't expose a `request.tag` variable, and `target.resource.tag` can't be evaluated when no target resource exists yet (per [OCI docs](https://docs.oracle.com/en-us/iaas/Content/Tagging/Tasks/managingaccesswithtags.htm)). Datadog code is the only consumer of this grant and always tags the resources it creates, so practical exposure is the same as tag-enforced create would be.

## Test plan

Validated end-to-end against `dddevintegration2` with the `dd-svc` user:

- **Event rule create with tag**: ✅ allowed (via `EVENTRULE_CREATE`)
- **Event rule create without tag**: ✅ allowed (same — Datadog won't actually do this)
- **Event rule update on tagged**: ✅ allowed
- **Event rule update on untagged**: ❌ `NotAuthorizedOrNotFound` (where-clause correctly scopes out)
- **Event rule delete on tagged / untagged**: ✅ / ❌
- **Stream create / update / delete**: same six-row matrix, all expected results
- **SCH stream-pull**: created two SCHes (source = tagged vs untagged stream), pushed a message to each:
  - Tagged-source SCH: `MessagesReadFromSource=1`, `MessagesWrittenToTarget=1`, errors=0
  - Untagged-source SCH: no metric datapoints emitted (the SCH service doesn't even attempt cycles when the dynamic-group identity is denied at the source)

## Backwards compatibility

Net diff is two new resources and four new policy statements per auth module — no changes to existing variables, schema, or resources. Stacks that don't actually use events still get these IAM additions but nothing exercises them.

### DIFF
--- a/datadog-integration/modules/auth/main.tf
+++ b/datadog-integration/modules/auth/main.tf
@@ -180,8 +180,22 @@ resource "oci_identity_domains_group" "dd_auth" {
   }
 }
 
+resource "oci_identity_tag_namespace" "datadog_managed" {
+  compartment_id = var.compartment_id
+  name           = "DatadogManaged"
+  description    = "[DO NOT REMOVE] Marker namespace for Datadog-managed resources. Removing this breaks the IAM grant scoping for the events rule."
+  freeform_tags  = var.tags
+  defined_tags   = var.defined_tags
+}
+
+resource "oci_identity_tag" "marker" {
+  tag_namespace_id = oci_identity_tag_namespace.datadog_managed.id
+  name             = "marker"
+  description      = "[DO NOT REMOVE] Applied to resources that the Datadog group is allowed to manage (e.g. the events rule). Removing this breaks the scoped IAM grant."
+}
+
 resource "oci_identity_policy" "dd_auth" {
-  depends_on     = [null_resource.user_group_variable_validation, oci_identity_domains_group.dd_auth]
+  depends_on     = [null_resource.user_group_variable_validation, oci_identity_domains_group.dd_auth, oci_identity_tag_namespace.datadog_managed]
   compartment_id = var.tenancy_id
   description    = "[DO NOT REMOVE] Policies required by Datadog User"
   name           = var.user_policy_name
@@ -194,7 +208,9 @@ resource "oci_identity_policy" "dd_auth" {
     "Allow group id ${local.dd_group_ocid} to manage buckets in compartment id ${var.compartment_id} where target.bucket.name=/dd-*/",
     "Allow group id ${local.dd_group_ocid} to manage object-family in compartment id ${var.compartment_id} where target.bucket.name=/dd-*/",
     "Allow group id ${local.dd_group_ocid} to use fn-invocation in compartment id ${var.compartment_id}",
-    "Endorse group id ${local.dd_group_ocid} to read objects in tenancy usage-report"
+    "Endorse group id ${local.dd_group_ocid} to read objects in tenancy usage-report",
+    "Allow group id ${local.dd_group_ocid} to manage cloudevents-rules in tenancy where any {request.permission = 'EVENTRULE_CREATE', target.resource.tag.DatadogManaged.marker = 'true'}",
+    "Allow group id ${local.dd_group_ocid} to manage streams in compartment id ${var.compartment_id} where any {request.permission = 'STREAM_CREATE', target.resource.tag.DatadogManaged.marker = 'true'}",
   ]
   freeform_tags = var.tags
   defined_tags  = var.defined_tags
@@ -229,7 +245,9 @@ resource "oci_identity_policy" "dynamic_group" {
     "Allow dynamic-group id ${oci_identity_domains_dynamic_resource_group.service_connector.ocid} to use fn-function in compartment id ${var.compartment_id}",
     "Allow dynamic-group id ${oci_identity_domains_dynamic_resource_group.service_connector.ocid} to use fn-invocation in compartment id ${var.compartment_id}",
     "Allow dynamic-group id ${oci_identity_domains_dynamic_resource_group.forwarding_function.ocid} to read secret-bundles in compartment id ${var.compartment_id}",
-    "Allow dynamic-group id ${oci_identity_domains_dynamic_resource_group.forwarding_function.ocid} to manage object-family in compartment id ${var.compartment_id} where target.bucket.name=/dd-*/"
+    "Allow dynamic-group id ${oci_identity_domains_dynamic_resource_group.forwarding_function.ocid} to manage object-family in compartment id ${var.compartment_id} where target.bucket.name=/dd-*/",
+    "Allow dynamic-group id ${oci_identity_domains_dynamic_resource_group.service_connector.ocid} to use stream-pull in compartment id ${var.compartment_id} where target.resource.tag.DatadogManaged.marker = 'true'",
+    "Allow any-user to use stream-push in compartment id ${var.compartment_id} where all {request.principal.type = 'eventrule', target.resource.tag.DatadogManaged.marker = 'true'}"
   ]
   freeform_tags = var.tags
   defined_tags  = var.defined_tags
--- a/datadog-terraform-onboarding/modules/auth/main.tf
+++ b/datadog-terraform-onboarding/modules/auth/main.tf
@@ -184,8 +184,22 @@ resource "oci_identity_domains_group" "dd_auth" {
   }
 }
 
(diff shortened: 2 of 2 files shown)


datadog-integration/modules/auth/main.tf
```diff
@@ -191,7 +213,18 @@ resource "oci_identity_policy" "dd_auth" {
...
-    "Endorse group id ${local.dd_group_ocid} to read objects in tenancy usage-report"
+    "Endorse group id ${local.dd_group_ocid} to read objects in tenancy usage-report",
+    # --- events ---
+    # HubManager toggles is_enabled on the events rule when the customer opts in/out via the Datadog UI.
+    # 'use' grants RULE_READ + RULE_UPDATE only; rule create/delete is done via the stack (customer admin).
+    "Allow group id ${local.dd_group_ocid} to use cloudevents-rules in tenancy where target.resource.tag.DatadogManaged.marker = 'true'",
+    # HubManager reads stream state (consumer lag, partition info). Stream lifecycle is owned by HubManager.
+    "Allow group id ${local.dd_group_ocid} to read stream-family in compartment id ${var.compartment_id}",
+    # --- backfill (added upfront to avoid customer re-apply when backfill ships) ---
+    "Allow group id ${local.dd_group_ocid} to manage buckets in compartment id ${var.compartment_id} where target.bucket.name=/dd-*-backfill-*/",
```
yuhuyoyo: i'm creating a tag, would you like to use tag as well? I think it's less fragil than name matching. @EamonBrady1

EamonBrady1: I think we already have a tag like ownedBy:Datadog or something. And wouldn't it be similarly fragile? I guess a user could change a resource name just as easily as a tag right?

EamonBrady1: Maybe we use the tag and then we set the permissions as "where (name=<expected_name> OR hasDatadogTag)"? That way we can use both?

yuhuyoyo: https://docs.oracle.com/en-us/iaas/Content/Tagging/Concepts/understandingfreeformtags.htm#:~:text=you%20can%27t%20include%20free%2Dform%20tags%20in%20IAM%20policies

unfortunately can't use freeform tag for iam policy so can't use the current ownedBy:Datadog.

I agree that user can modify all the datadog resource if they want so fine with me to keep dd-*

datadog-integration/modules/auth/main.tf
```diff
@@ -194,7 +208,18 @@ resource "oci_identity_policy" "dd_auth" {
     "Allow group id ${local.dd_group_ocid} to manage buckets in compartment id ${var.compartment_id} where target.bucket.name=/dd-*/",
     "Allow group id ${local.dd_group_ocid} to manage object-family in compartment id ${var.compartment_id} where target.bucket.name=/dd-*/",
     "Allow group id ${local.dd_group_ocid} to use fn-invocation in compartment id ${var.compartment_id}",
-    "Endorse group id ${local.dd_group_ocid} to read objects in tenancy usage-report"
+    "Endorse group id ${local.dd_group_ocid} to read objects in tenancy usage-report",
+    # HubManager creates/updates/deletes the events rule when the customer opts in/out via the Datadog UI.
+    # CREATE is allowed unconditionally (OCI cannot evaluate target.resource.tag on creates since the resource
+    # doesn't exist yet); UPDATE/DELETE/READ are scoped to rules carrying the DatadogManaged.marker tag so
+    # HubManager cannot modify or delete unrelated rules. HubManager is the only consumer of this grant and
```
EamonBrady1: We can probably delete these comments. Since this is public code probably best not to mention things like hubmanager

datadog-integration/modules/auth/main.tf
```diff
@@ -194,7 +208,18 @@ resource "oci_identity_policy" "dd_auth" {
...
+    # always tags rules it creates.
+    "Allow group id ${local.dd_group_ocid} to manage cloudevents-rules in tenancy where any {request.permission = 'EVENTRULE_CREATE', target.resource.tag.DatadogManaged.marker = 'true'}",
+    # HubManager creates the events stream in the Datadog compartment (and tears it down on opt-out).
+    # CREATE is allowed unconditionally (target.resource.tag can't be evaluated on creates);
+    # READ/UPDATE/DELETE are scoped to streams carrying the DatadogManaged.marker tag.
+    # OCI streaming has no target.stream.name variable, so tag-based scoping is the only narrow option.
```
EamonBrady1: same with these

EamonBrady1 (approved):
LGTM. Can we remove the comments above the policies though?

datadog-integration/modules/auth/main.tf
yuhuyoyo: Good catch — rewrote the comments to drop the internal service name. The two block-comments and the resource descriptions now refer to it as 'the Datadog group' / 'Datadog-owned' etc. Also scrubbed the commit message and PR body. Fix in commit `9fe63f5`.

datadog-integration/modules/auth/main.tf
yuhuyoyo: Same scrub applied here, in the dynamic_group block, and across the diff — `grep -i hubmanager` returns nothing now.

datadog-integration/modules/auth/main.tf
EamonBrady1: Sweet, thanks! Up to you on if you want to keep the other comments, I'm good with either way

yuhuyoyo: Dropped them in `c44a82e` — the policy statements are self-describing enough. Net PR diff is now +42/-6 in the two auth modules.

PR #126, opened 2026-05-12
refactor: extract dd_group_ocid local in auth module

yuhuyoyo:
## What

Extracts a `dd_group_ocid` local in `modules/auth/locals.tf` to replace five repeated inline ternaries in the IAM policy statements:

```hcl
# before — repeated 5×
"Allow group id ${var.existing_group_id != null && var.existing_group_id != "" ? var.existing_group_id : oci_identity_domains_group.dd_auth[0].ocid} to ..."

# after
"Allow group id ${local.dd_group_ocid} to ..."
```

## Why

Pure readability refactor. The policy statements were hard to scan because the group OCID resolution logic was inlined five times. No behaviour change.

## Stacking

ECI-1615 (#125) is based on this branch.

### DIFF
--- a/datadog-integration/modules/auth/locals.tf
+++ b/datadog-integration/modules/auth/locals.tf
@@ -1,3 +1,4 @@
 locals {
-  email = var.user_email
+  email         = var.user_email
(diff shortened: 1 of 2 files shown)


EamonBrady1 (approved):
Niceeee

PR #133, opened 2026-05-19
ECI-1542: Add logs_only onboarding variable

EamonBrady1:
For reference: [Datadog Logs Only Onboarding Flow
](https://app.datadoghq.com/signup/setup/logs/cloud-provider)
## What
Adds a `logs_only` bool variable to the `datadog-integration` (OCI ORM Stack) and `datadog-terraform-onboarding` (advanced Terraform) modules. When `logs_only=true`, the create request to the Datadog OCI integration API sends `resource_collection_enabled=false` and `metrics_config.enabled=false`, so the tenancy is onboarded for logs collection only. Updates made to these fields from stack requests are ignored, similar to how logs behave.

## Why
[ECI-1542](https://datadoghq.atlassian.net/browse/ECI-1542) — Support a logs-only onboarding flow for OCI from the Datadog UI (`/signup/setup/logs/cloud-provider`). The signup UI will deep-link into OCI Resource Manager with `logs_only:true`, and the stack must propagate that into the integration request so the org isn't billed for metrics/resources it didn't ask for. The companion dd-source PR makes the API ignore subsequent metrics/resource updates from these Terraform clients so users can later enable them in the UI without the stack reverting them.

## Testing
- `terraform validate` passes for both `datadog-integration` and `datadog-terraform-onboarding`.
- Inspected the rendered `request_data` JSON: with `logs_only=false` the payload is unchanged (no `resource_collection_enabled` or `metrics_config` keys); with `logs_only=true` both fields are present with disabled values.
- Default of `false` preserves existing behavior for current users.

[ECI-1542]: https://datadoghq.atlassian.net/browse/ECI-1542?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/datadog-integration/modules/integration/locals.tf
+++ b/datadog-integration/modules/integration/locals.tf
@@ -1,27 +1,34 @@
 locals {
 
   config_version = 2
+  base_attributes = {
+    home_region : var.home_region
+    user_ocid : var.user_ocid
+    config_version : local.config_version
+    auth_credentials : {
+      private_key : var.private_key
+    },
+    regions_config : {
+      available : var.subscribed_regions
+    }
+    dd_compartment_id : var.datadog_resource_compartment_id
+    dd_stack_id : try(data.external.stack_info.result.stack_id, "")
+    logs_config : {
+      Enabled : var.logs_enabled
+    }
+    defined_tags : [for k, v in var.defined_tags : "${k}:${v}"]
+  }
+  logs_only_attributes = { for k, v in {
+    resource_collection_enabled : false
+    metrics_config : {
+      enabled : false
+    }
+  } : k => v if var.logs_only }
   json_object = {
     data : {
       type : "oci_tenancy",
       id : var.tenancy_ocid,
-      attributes : {
-        home_region : var.home_region
-        user_ocid : var.user_ocid
-        config_version : local.config_version
-        auth_credentials : {
-          private_key : var.private_key
-        },
-        regions_config : {
-          available : var.subscribed_regions
-        }
-        dd_compartment_id : var.datadog_resource_compartment_id
-        dd_stack_id : try(data.external.stack_info.result.stack_id, "")
-        logs_config : {
-          Enabled : var.logs_enabled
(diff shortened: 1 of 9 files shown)


datadog-integration/modules/integration/locals.tf
```diff
@@ -1,27 +1,34 @@
 locals {
 
   config_version = 2
```
sva91, line 3: should we update the `config version` here ?

EamonBrady1: I don't think that's necessary, do you? Historically we've kept it at 2 for quickstart, even when we added logs_config to the API object, so I don't think adding the metrics or resource collection configs requires a version change, especially since it won't impact any existing customers, or induce new behavior for new customers besides deciding what is/isn't enabled.

sva91, line 3: If there is an accompanying PR for dd-source, let me know just so that I have full context on this change. This isn't a blocker though.

sva91 (approved):

datadog-integration/modules/integration/locals.tf
EamonBrady1: Yep I've got it covered, sorry I meant to link this in the PR description: https://github.com/DataDog/dd-source/pull/442816

sva91: ok, we are making changes to stack which requires changes in more than one service. each of them should have been a config version update in my opinion. we can talk about this outside the scope of your current work though.

EamonBrady1: Definitely agreed on that. I think we have missed many opportunities to update our versioning, even just to something like 2.x, so we could more specifically tie bugs to changes made to the stack. Could be worth having the discussion about how we want to address this moving forward and whether we want to just start incrementing it in the future

PR #136, opened 2026-06-02
ECI-1701: Poll vault management endpoint to avoid DNS race condition on KMS key creation

EamonBrady1:
## What
Adds a \`null_resource\` polling loop between \`oci_kms_vault\` and \`oci_kms_key\` creation in both the \`datadog-integration\` and \`datadog-terraform-onboarding\` KMS modules. Also adds the \`hashicorp/null\` provider to both KMS module \`required_providers\` blocks.

## Why
Customers hit a known race condition in the OCI Terraform provider: after a KMS vault is marked active, its management endpoint DNS is not immediately resolvable in all regions, causing the subsequent \`oci_kms_key\` creation to fail.

## Testing
Heavily tested to ensure the race condition is fully addressed. The polling loop probes the vault's own management endpoint (the exact DNS name the OCI provider uses to create the key) via the OCI CLI every 10 seconds, exiting as soon as it gets a successful response. A fixed \`time_sleep\` of 60s was first attempted but proved insufficient. The polling approach was validated to correctly detect vault reachability and unblock key creation. Times out after 300s with a message to re-apply.

### DIFF
--- a/datadog-terraform-onboarding/providers.tf
+++ b/datadog-terraform-onboarding/providers.tf
@@ -17,6 +17,10 @@ terraform {
       source  = "hashicorp/null"
       version = ">= 3.0"
     }
+    time = {
+      source  = "hashicorp/time"
+      version = ">= 0.9.0"
+    }
   }
 }
 
--- a/datadog-terraform-onboarding/providers.tf
+++ b/datadog-terraform-onboarding/providers.tf
@@ -17,10 +17,6 @@ terraform {
       source  = "hashicorp/null"
       version = ">= 3.0"
     }
-    time = {
-      source  = "hashicorp/time"
-      version = ">= 0.9.0"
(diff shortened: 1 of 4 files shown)


datadog-terraform-onboarding/providers.tf
```diff
@@ -17,6 +17,10 @@ terraform {
       source  = "hashicorp/null"
       version = ">= 3.0"
     }
+    time = {
```
bengillmandd: Is this needed? I don't see it used in this pr

bengillmandd (approved):
lgtm just one small comment

datadog-terraform-onboarding/providers.tf
EamonBrady1: Good call, dropped it

PR #137, opened 2026-06-08
fix(events-forwarder): decode SCH streaming base64 envelopes, simplify Decode

yuhuyoyo:
## Summary

- Decode SCH OCI Streaming envelopes by base64-decoding each message's `value` field before forwarding to Datadog
- Simplify `Decode` to handle only SCH OCI Streaming format (single JSON pass into `[]streamingMessage`), removing the `isStreamingFormat` heuristic and multi-path fallback
- Update handler and formatter tests to use SCH-wrapped payloads via a new `schBatch` helper

## Test plan

- [ ] `go test ./...` passes in `datadog-functions/events-forwarder`
- [ ] `TestDecode_SCHStreamingArray` / `TestDecode_SCHStreamingSingleMessage` verify base64 unwrapping
- [ ] `TestMyHandler_ArrayPayload` / `TestMyHandler_SingleEnvelope` verify end-to-end handler with SCH format
- [ ] `TestMyHandler_MissingTenancyOCID` / `TestMyHandler_MissingDDSite` correctly reach env-var validation

### DIFF
--- a/datadog-functions/events-forwarder/internal/formatter/formatter.go
+++ b/datadog-functions/events-forwarder/internal/formatter/formatter.go
@@ -34,6 +34,7 @@ func Decode(in io.Reader) ([]json.RawMessage, error) {
 	}
 	out := make([]json.RawMessage, 0, len(msgs))
 	for i, sm := range msgs {
+		// OCI Streaming encodes message values as standard RFC 4648 base64 (with padding).
 		decoded, err := base64.StdEncoding.DecodeString(sm.Value)
 		if err != nil {
 			return nil, fmt.Errorf("streaming message %d: failed to base64-decode value: %w", i, err)
@@ -43,6 +44,9 @@ func Decode(in io.Reader) ([]json.RawMessage, error) {
 		}
 		if decoded[0] != '{' {
 			return nil, fmt.Errorf("streaming message %d: decoded value is not a JSON object", i)
+		}
+		if !json.Valid(decoded) {
+			return nil, fmt.Errorf("streaming message %d: decoded value is not valid JSON", i)
 		}
 		out = append(out, json.RawMessage(decoded))
 	}
--- a/datadog-functions/events-forwarder/internal/formatter/formatter.go
+++ b/datadog-functions/events-forwarder/internal/formatter/formatter.go
@@ -41,8 +41,8 @@ func Decode(in io.Reader) ([]json.RawMessage, error) {
 		if len(decoded) == 0 {
 			return nil, fmt.Errorf("streaming message %d: decoded value is empty", i)
 		}
-		if !json.Valid(decoded) {
-			return nil, fmt.Errorf("streaming message %d: decoded value is not valid JSON", i)
+		if decoded[0] != '{' {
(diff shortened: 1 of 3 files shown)


datadog-functions/events-forwarder/internal/formatter/formatter.go
```diff
@@ -4,36 +4,43 @@
...
-	var obj map[string]any
-	if err := json.Unmarshal(body, &obj); err != nil {
-		return nil, fmt.Errorf("invalid JSON format: expected object or array of objects: %w", err)
+	out := make([]json.RawMessage, 0, len(msgs))
+	for i, sm := range msgs {
+		decoded, err := base64.StdEncoding.DecodeString(sm.Value)
+		if err != nil {
+			return nil, fmt.Errorf("streaming message %d: failed to base64-decode value: %w", i, err)
+		}
+		out = append(out, json.RawMessage(decoded))
```
rheei, line 51: `decoded` is appended as `json.RawMessage` without any validity check. Two paths reach this silently:

1. **Non-JSON value** — if a message's base64 decodes to non-JSON bytes (truncated event, wrong encoding), the error surfaces later inside `Chunk`'s `json.Marshal` flush with no message index and no hint of which event was bad.
2. **Empty value** — a missing or `""` `value` field: `base64.StdEncoding.DecodeString("")` returns `([]byte{}, nil)`, so a zero-length `RawMessage` is appended and `Chunk` fails with `unexpected end of JSON input`.

Suggest adding before the append:
```go
if len(decoded) == 0 {
    return nil, fmt.Errorf("streaming message %d: decoded value is empty", i)
}
if !json.Valid(decoded) {
    return nil, fmt.Errorf("streaming message %d: decoded value is not valid JSON", i)
}
```

— Claude (via @bella.rhee)

datadog-functions/events-forwarder/internal/formatter/formatter.go
```diff
@@ -4,36 +4,49 @@
...
+	out := make([]json.RawMessage, 0, len(msgs))
+	for i, sm := range msgs {
+		decoded, err := base64.StdEncoding.DecodeString(sm.Value)
+		if err != nil {
+			return nil, fmt.Errorf("streaming message %d: failed to base64-decode value: %w", i, err)
+		}
+		if len(decoded) == 0 {
+			return nil, fmt.Errorf("streaming message %d: decoded value is empty", i)
+		}
+		if !json.Valid(decoded) {
```
rheei: `json.Valid` accepts any syntactically valid JSON — scalars like `null`, `42`, or `"str"` all return true. OCI Cloud Events are always objects, so a corrupted or mis-encoded message whose base64 decodes to a scalar would pass this guard, enter `Chunk`, and cause the intake to reject the whole batch silently.

Since `len(decoded) > 0` is already guaranteed by the guard above, `decoded[0]` is safe to check:

```go
if decoded[0] != '{"' {
    return nil, fmt.Errorf("streaming message %d: decoded value is not a JSON object", i)
}
```

This is the idiomatic Go pattern here — zero allocation, and clearer intent than `json.Valid` alone.

— Claude (via @bella.rhee)

rheei (approved):

PR #142, opened 2026-06-17
ECI-1658 [OCI] detect 5xx errors and attempt to write to OCI bucket

erics-ddog:
**what:**
Added a condition in the `sendMessage` function of `client.go` to catch 5xx responses, and then attempt to write those data points to an OCI bucket.

(Bucket creation and replay will be in future PRs).

**why:**
During Datadog outages (which produce the 5xx responses), all data points received from OCI are currently lost. We instead write these points to an OCI bucket, so we can avoid losing them and backfill later (backfill is a future implementation).

**testing:**
Manual testing involved ensuring that 2xx responses still behave as expected. We then simulated a 5xx response by rerouting the 2xx responses to trigger the filling of buckets. Both cases passes their respective tests, behaving as desired.


datadog-functions/lib/client/client.go
```diff
@@ -136,3 +150,123 @@ func NewDatadogClientWithTenancyAndSite() (DatadogClient, string, string, error)
...
+func backfillBucketName(intakeURL string) string {
+	switch {
+	case strings.Contains(intakeURL, "ocimetrics"):
+		return "dd-metrics-backfill"
+	case strings.Contains(intakeURL, "cloudchanges"):
+		return "dd-events-backfill"
+	case strings.Contains(intakeURL, "logs"):
+		return "dd-logs-backfill"
+	default:
+		return ""
```
cale-at-ddog: Nit: Did you consider returning an error here, rather than relying on an empty string? I'm not sure which is more idiomatic for a string-return-type function like this.

datadog-functions/lib/client/client.go
```diff
@@ -136,3 +150,123 @@ func NewDatadogClientWithTenancyAndSite() (DatadogClient, string, string, error)
...
+
+// handleServerErrorPayload persists a payload that Datadog rejected with a 5xx to
+// a per-data-type backfill bucket so it can be replayed later. The bucket is
+// expected to already exist (provisioned out-of-band); if it does not, or the
+// data type cannot be determined, the payload is dropped (logged). This is
+// best-effort and never changes the status/error returned to the caller.
+func handleServerErrorPayload(ctx context.Context, message []byte, intakeURL string) {
+	bucket := backfillBucketName(intakeURL)
+	if bucket == "" {
+		log.Printf("5xx payload dropped: could not determine data type from url %q", intakeURL)
```
cale-at-ddog: Do we want to attach a log level to these messages?

datadog-functions/lib/client/client.go
```diff
@@ -136,3 +150,123 @@ func NewDatadogClientWithTenancyAndSite() (DatadogClient, string, string, error)
...
+		return
+	}
+
+	osClient, err := newObjectStorageClient()
+	if err != nil {
+		log.Printf("5xx payload dropped: %v", err)
```
cale-at-ddog: It might be worth adding a clause to this error that it's a storage-client error (like the clauses in the other messages identifying where they come from).

datadog-functions/lib/client/client.go
```diff
@@ -1,16 +1,22 @@
 package client
```
bengillmandd, line 1: Is it worth adding unit tests for any functionality introduced in this PR?

datadog-functions/lib/client/client.go
erics-ddog: I have to do more research on how to do OCI log capture since it doesn't seem to be something we are currently doing. This is more involved than what this PR is meant for – will make a new ticket for it.

datadog-functions/lib/client/client.go
erics-ddog: Added some tests in latest commit

bengillmandd (approved):

PR #146, opened 2026-06-23
ECI-1815: add version-bump workflow for terraform module changes

yuhuyoyo:
## What

Adds `.github/workflows/version-bump.yml` — a dedicated workflow that automatically bumps the patch version in both `VERSION` (root) and `datadog-terraform-onboarding/VERSION` whenever a PR touches `datadog-integration/` or `datadog-terraform-onboarding/`.

The `paths:` trigger ensures the workflow only fires for relevant changes; PRs that touch only Go code or docs never see this workflow run.

**Behaviour:**
- Reads the base-branch `VERSION` via the GitHub API, increments the patch digit, writes the new value to both VERSION files.
- Commits the bump directly to the PR branch using `GITHUB_TOKEN` (these commits do not re-trigger workflows).
- Posts a PR comment with the before/after versions and guidance for manual major/minor bumps.
- Skips if the PR's `VERSION` is already ahead of the base (idempotent on re-pushes and after manual bumps).
- Only runs for same-repository PRs (fork PRs excluded — `GITHUB_TOKEN` cannot push to a fork branch).

## Why

Manual version bumps were easy to forget, and the release workflow already enforces that both VERSION files must match before cutting a release. Automating the patch bump removes that friction while keeping major/minor bump decisions in human hands.

See #145 for the proof-of-concept that verified the approach works.

## Test plan

- [ ] Merge this PR, then open a test PR with any change under `datadog-integration/` or `datadog-terraform-onboarding/` and verify the bump fires.
- [ ] Push a second commit to that PR and verify the workflow skips (already-bumped detection).
- [ ] Manually set a higher version in the PR and verify the workflow skips again.
- [ ] Open a PR that only touches Go code and verify this workflow does not appear in the Checks tab.


.github/workflows/version-bump.yml
```diff
@@ -0,0 +1,111 @@
...
+          pr_version=$(tr -d '[:space:]' < VERSION)
+          pr_tf_version=$(tr -d '[:space:]' < datadog-terraform-onboarding/VERSION)
+          base_version=$(gh api "repos/${REPO}/contents/VERSION?ref=${BASE_REF}" \
+            --jq '.content' | base64 -d | tr -d '[:space:]')
+
+          IFS='.' read -r major minor patch <<< "$base_version"
+          bumped_version="${major}.${minor}.$((patch + 1))"
+
+          version_int() { IFS='.' read -r M m p <<< "$1"; echo $((M * 1000000 + m * 1000 + p)); }
+          base_int=$(version_int "$base_version")
```
rheei: **[HIGH]** `bump_tf` compares `pr_tf_version` against root's `base_int`, not TF's own base version. When a developer manually bumps root `VERSION=2.0.0` (breaking change) but leaves `datadog-terraform-onboarding/VERSION` at `1.5.2`, this sets `bump_tf=true` and writes `bumped_version` (= `base_patch+1` = `1.5.3`) to TF — leaving root=`2.0.0`, TF=`1.5.3`. The release workflow that enforces both files match then fails.

Fix: when `bump_root=false`, use `pr_version` as the target for TF:
```bash
bump_target_tf="$bumped_version"
[[ "$bump_root" == "false" ]] && bump_target_tf="$pr_version"
```
Then write `$bump_target_tf` to `datadog-terraform-onboarding/VERSION` instead of `$bumped_version`.

— Claude (via @bella.rhee)

.github/workflows/version-bump.yml
```diff
@@ -0,0 +1,111 @@
...
+      - name: Compute versions
+        id: versions
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          BASE_REF: ${{ github.base_ref }}
+          REPO: ${{ github.repository }}
+        run: |
+          pr_version=$(tr -d '[:space:]' < VERSION)
+          pr_tf_version=$(tr -d '[:space:]' < datadog-terraform-onboarding/VERSION)
+          base_version=$(gh api "repos/${REPO}/contents/VERSION?ref=${BASE_REF}" \
```
rheei: **[MEDIUM]** Two concurrent PRs touching `datadog-integration/` against the same base both compute the same `bumped_version` (e.g. both get `1.5.3` from base `1.5.2`). If both merge without the second rebasing, main ends up at `1.5.3` twice — two distinct code changes shipping under the same version tag. Only mitigated if "require branches to be up to date" branch protection is enabled. Worth documenting as a known limitation.

— Claude (via @bella.rhee)

.github/workflows/version-bump.yml
```diff
@@ -0,0 +1,111 @@
...
+          [[ "$BUMP_ROOT" == "true" ]] && printf '%s\n' "$BUMPED_VERSION" > VERSION
+          [[ "$BUMP_TF"   == "true" ]] && printf '%s\n' "$BUMPED_VERSION" > datadog-terraform-onboarding/VERSION
+
+      - name: Commit and push
+        if: steps.versions.outputs.skip == 'false'
+        env:
+          BASE_VERSION: ${{ steps.versions.outputs.base_version }}
+          BUMPED_VERSION: ${{ steps.versions.outputs.bumped_version }}
+        run: |
+          git config user.name "github-actions[bot]"
```
rheei, line 49: **[MEDIUM]** `git push` with no refspec can fail non-fast-forward if the PR author pushes between this workflow's checkout and this push. `cancel-in-progress: true` reduces the window but doesn't close it — if the author's push lands just as this workflow executes `git push`, it gets rejected and the bump doesn't land (leaving a spurious red check; the next push retries correctly).

Consider rebasing before pushing:
```bash
git pull --rebase origin ${{ github.head_ref }}
git push
```

— Claude (via @bella.rhee)

.github/workflows/version-bump.yml
```diff
@@ -0,0 +1,135 @@
...
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git add VERSION datadog-terraform-onboarding/VERSION
+          git commit -m "chore: bump version ${BASE_VERSION} -> ${BUMPED_VERSION}"
+          git pull --rebase origin ${{ github.head_ref }}
```
rheei: **[HIGH] Shell injection regression** — `${{ github.head_ref }}` is interpolated directly into the shell script before bash sees it. A specially-crafted branch name containing shell metacharacters (`;`, `&&`, `$()`) executes arbitrary commands in the runner (which holds `contents: write` + `pull-requests: write`).

Commit 2 in this PR explicitly fixed this pattern everywhere: _"Pass all step outputs through env vars (never inline `${{ }}` in run: scripts) to close the shell injection vector."_ This line reintroduces it.

Fix: add `HEAD_REF` to the step's `env:` block and use `$HEAD_REF` in the shell:
```yaml
env:
  BASE_VERSION: ${{ steps.versions.outputs.base_version }}
  BUMPED_VERSION: ${{ steps.versions.outputs.bumped_version }}
  HEAD_REF: ${{ github.head_ref }}
```

— Claude (via @bella.rhee)

.github/workflows/version-bump.yml
```diff
@@ -0,0 +1,135 @@
...
+        run: |
+          pr_version=$(tr -d '[:space:]' < VERSION)
+          pr_tf_version=$(tr -d '[:space:]' < datadog-terraform-onboarding/VERSION)
+          base_version=$(gh api "repos/${REPO}/contents/VERSION?ref=${BASE_REF}" \
+            --jq '.content' | base64 -d | tr -d '[:space:]')
+          base_tf_version=$(gh api "repos/${REPO}/contents/datadog-terraform-onboarding/VERSION?ref=${BASE_REF}" \
+            --jq '.content' | base64 -d | tr -d '[:space:]')
+
+          IFS='.' read -r major minor patch <<< "$base_version"
+          auto_bumped_version="${major}.${minor}.$((patch + 1))"
```
rheei: **[HIGH] `auto_bumped_version` ignores TF's own base — can regress TF if the two files have diverged**

`auto_bumped_version` is always `base_version_patch + 1` (root only). `base_tf_version` is fetched and used for manual-bump detection but never sets a floor for the auto-bump target.

If base root=`1.5.2` and base TF=`1.5.5` (legitimate historical divergence), neither file is manually bumped on the PR branch, so `target_version = 1.5.3`. `pr_tf_version (1.5.5) != 1.5.3` → `bump_tf=true` → TF is written to `1.5.3`, a regression from `1.5.5`.

Fix: use `max(auto_bumped_version, base_tf_patch + 1)` as the auto-bump target, or assert both base files are equal before proceeding and fail fast if they aren't.

— Claude (via @bella.rhee)

.github/workflows/version-bump.yml
rheei: **[MEDIUM] Commit-then-rebase ordering leaves the runner in unrecoverable REBASE_HEAD state on conflict**

The sequence is: commit locally → `git pull --rebase` → push. If the remote has a conflicting change to `VERSION` (e.g. from a concurrent merge), the rebase exits with a conflict, leaving the job mid-rebase with a confusing error and no recovery path.

Safe ordering: pull --rebase _before_ writing files, then commit, then push. Requires moving the rebase to before the Write step but eliminates the conflict window entirely.

— Claude (via @bella.rhee)

.github/workflows/version-bump.yml
```diff
@@ -0,0 +1,135 @@
...
+          # Detect valid manual bumps (strictly above the respective base).
+          root_manually_bumped=false
+          tf_manually_bumped=false
+          [[ $(version_int "$pr_version")    -gt $base_int    ]] && root_manually_bumped=true
+          [[ $(version_int "$pr_tf_version") -gt $base_tf_int ]] && tf_manually_bumped=true
+
+          # If a developer manually bumped one file, propagate that version to
+          # both files so they stay in sync. Root takes priority over TF when
+          # both were bumped to different values.
+          if [[ "$root_manually_bumped" == "true" ]]; then
```
rheei: **[MEDIUM] When both files are manually bumped to different values, TF's value is silently discarded**

If a developer bumps root `VERSION` to `2.0.0` and `datadog-terraform-onboarding/VERSION` to `1.6.0` for separate reasons, `root_manually_bumped=true` wins the `if/elif`, `target_version=2.0.0`, and TF is silently overwritten to `2.0.0`. The developer's explicit `1.6.0` for TF is lost without warning.

At minimum, log when this happens:
```bash
if [[ "$root_manually_bumped" == "true" && "$tf_manually_bumped" == "true" && "$pr_version" != "$pr_tf_version" ]]; then
  echo "WARNING: both files bumped to different values ($pr_version vs $pr_tf_version); using root for both."
fi
```

— Claude (via @bella.rhee)

.github/workflows/version-bump.yml
```diff
@@ -0,0 +1,77 @@
...
+        if: steps.versions.outputs.skip == 'false'
+        env:
+          BASE_VERSION: ${{ steps.versions.outputs.base_version }}
+          BUMPED_VERSION: ${{ steps.versions.outputs.bumped_version }}
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git add VERSION datadog-terraform-onboarding/VERSION
+          git commit -m "chore: bump version ${BASE_VERSION} -> ${BUMPED_VERSION}"
+          git pull --rebase origin ${{ github.head_ref }}
```
rheei: **[HIGH] Shell injection still present** — `${{ github.head_ref }}` is interpolated directly into the shell script before bash sees it. A branch name containing shell metacharacters (`;`, `&&`, `$()`) executes arbitrary commands in the runner.

This was flagged in the previous round. The `Compute and write versions` step correctly gates everything behind `env:` vars; this step regresses that pattern. Fix: add `HEAD_REF: ${{ github.head_ref }}` to the step's `env:` block and use `$HEAD_REF` in the shell.

— Claude (via @bella.rhee)

rheei: **[MEDIUM] Commit-then-rebase ordering** — the local version-bump commit is created before `git pull --rebase`, so if the remote branch has advanced since checkout (concurrent push), the rebase replays the bump commit on top. If both sides touched `VERSION`, rebase conflicts abort with no recovery path in CI. Safe order: pull --rebase first (before writing files), then commit, then push.

— Claude (via @bella.rhee)

scripts/bump-version.sh
```diff
@@ -0,0 +1,91 @@
...
+fi
+
+if [[ -n "${BASE_TF_VERSION_OVERRIDE:-}" ]]; then
+  base_tf_version="$BASE_TF_VERSION_OVERRIDE"
+else
+  base_tf_version=$(gh api "repos/${REPO}/contents/datadog-terraform-onboarding/VERSION?ref=${BASE_REF}" \
+    --jq '.content' | base64 -d | tr -d '[:space:]')
+fi
+
+IFS='.' read -r major minor patch <<< "$base_version"
```
rheei, line 41: **[HIGH] `auto_bumped_version` ignores `base_tf_version` — can regress TF if the two base files have diverged**

`auto_bumped_version` is always `base_root_patch + 1`. `base_tf_version` is fetched and used for manual-bump detection but never sets a floor for the auto-bump target.

If base root=`1.5.2` and base TF=`1.5.5` (legitimate historical divergence), neither file is manually bumped on the PR branch, so `target_version = 1.5.3`. `pr_tf_version (1.5.5) != 1.5.3` → `bump_tf=true` → TF is written to `1.5.3`, a regression from `1.5.5`.

Fix: floor the auto-bump at each file's own base:
```bash
IFS='.' read -r tf_major tf_minor tf_patch <<< "$base_tf_version"
auto_bumped_tf_version="${tf_major}.${tf_minor}.$((tf_patch + 1))"
# then use max(auto_bumped_version, auto_bumped_tf_version) as the target
```
Or: assert both base files are equal on entry and fail fast if they aren't.

— Claude (via @bella.rhee)

scripts/test-bump-version.bats
```diff
@@ -0,0 +1,95 @@
...
+  run_bump  "1.5.2" "1.5.2"
+  [ "$status" -eq 0 ]
+  [ "$(field bumped_version)" = "2.0.0"  ]
+  [ "$(field bump_root)"      = "false"  ]
+  [ "$(field bump_tf)"        = "true"   ]
+  [ "$(field skip)"           = "false"  ]
+}
+
+@test "TF bumped above its own base while root stays at base" {
+  make_tree "1.2.0" "1.2.1"
```
rheei, line 90: **[MEDIUM] Missing test: base TF ahead of base root, neither file manually bumped**

The regression scenario introduced by the `auto_bumped_version` issue above has no test coverage:

```bash
@test "base TF ahead of base root, neither manually bumped: TF must not regress" {
  make_tree "1.5.2" "1.5.5"   # PR files == their bases (no manual bump)
  run_bump  "1.5.2" "1.5.5"
  [ "$status" -eq 0 ]
  # Currently FAILS: writes 1.5.3 to TF, regressing from 1.5.5
  [ "$(field bumped_version)" = "1.5.6" ]  # or at minimum > 1.5.5
}
```

Adding this test would make the bug in `bump-version.sh` line 36 immediately visible.

— Claude (via @bella.rhee)

.github/workflows/version-bump.yml
yuhuyoyo: the two version files should always be the same. the release workflow fails if they are different https://github.com/DataDog/oracle-cloud-integration/blob/master/.github/workflows/release.yml#L26. so I just updated this to fail early if they are different and requires human to pick a correct version.

.github/workflows/version-bump.yml
yuhuyoyo: there can only be one concurrent run for this workflow per PR so the race should be quiet rare as a new commit will cancel out the old workflow. 

on an unfortunate case when a user push a new commit while this is running to the version file, the workflow will fail and a re-run will succeed as a noop.

.github/workflows/version-bump.yml
yuhuyoyo: this is by design the version must stay the same. we can log a warning tho

rheei (approved):
assuming this has been tested

.github/workflows/version-bump.yml
yuhuyoyo: i think claude is making duplicate comments?

scripts/bump-version.sh
yuhuyoyo: ditto

scripts/test-bump-version.bats
yuhuyoyo: done

PR #147, opened 2026-06-24
ECI-1659 [OCI] Add backfill mode to forwarder functions

erics-ddog:
**what:**

Adds a "backfill mode" to the three OCI forwarder functions (logs, metrics, events). When a forwarder is invoked with the control message `{"backfill_mode":"true"}` (instead of normal telemetry), it drains its region's backfill bucket back to Datadog rather than forwarding new data.

- **Detection:** `MyHandler` in each forwarder reads the invocation input once, and if it is the backfill trigger calls the shared `DatadogClient.Backfill(...)`; otherwise it runs the normal forward path from the buffered bytes (events also passes its `Dd-Oci-Tenancy-Id` header).
- **Drain loop (`Backfill`)**: paginates the per-region/per-type bucket (`dd-logs-backfill` / `dd-metrics-backfill` / `dd-events-backfill`), and for each object: `GetObject` → replay to Datadog → `DeleteObject` on success. Objects are stored already-batched and ready to send, so replay is just resending the bytes.
- **Replay-aware send:** the send core was split so the forward path (`SendMessageToDatadog`) still persists to the bucket on a 5xx, while the replay path (`replaySend`) does **not** re-bucket on failure — a failed replay is simply left in place for the next run. This avoids duplicating objects in the bucket.
- **Retry / throttle policy:** a replay is retried once; on a 429 it waits 5s before the retry; a second failure stops the run (the object is left for the next invocation).

**why:**

ECI-1658 made the forwarders persist undeliverable batches to a backfill bucket on a Datadog 5xx instead of dropping them. This PR is the consumer side: it replays those batches once Datadog is healthy., making the pipeline self-healing. Backfill is intended to be driven on a recurring (~5 min) schedule, so the function deliberately does no internal "graceful stop" or per-run cap — it drains until the bucket is empty, an unrecoverable failure occurs, or it times out, all of which are safe to be interrupted by because undelivered objects are left in place and the next run picks them up. Delivery is at-least-once (a send that succeeds but whose delete fails is re-sent next run).

**testing:**

- `go build` / `go vet` / `go test` pass for `lib` and all three forwarder modules; `gofmt` clean.
- Unit tests (`lib/client/backfill_test.go`):
  - `TestSendMessageBackfillTriggeredOnlyOn5xx` — the forward path persists to the bucket only on 5xx (not 2xx/4xx).
  - `TestBackfill` — replays and deletes each object on success; leaves the object and stops the run when a send fails after retry; waits-then-retries on a 429; empty bucket is a no-op; unrecognized intake URL errors. The Datadog mock now records `SentBodies`, so a test asserts the exact object bytes read from the bucket reach the Datadog send (and that replay never re-buckets).
  - `TestHandleServerErrorPayload` — the existing 5xx write-path coverage still passes.
- Object Storage is mocked via the `objectStorageAPI` seam (a fake `ListObjects`/`GetObject`/`DeleteObject`/`PutObject`); Datadog sends are mocked via the existing `MockAPIClient`. No real OCI/Datadog calls in tests.
- Manual testing: created a `dd-<type>-backfill` bucket in OCI, rerouted data points (2xx message) to said bucket, and invoked the forwarder with `{"backfill_mode":"true"}`. The payloads successfully land in Datadog and are deleted from the bucket.

**notes:**

Out of scope: function timeouts and creation of the `dd-*-backfill` buckets (all done by hubmanager). The events-forwarder handler code is included here so its image is ready; backfill is a no-op until the buckets exist.

### DIFF
--- a/datadog-functions/events-forwarder/internal/handler/handler.go
+++ b/datadog-functions/events-forwarder/internal/handler/handler.go
@@ -1,6 +1,7 @@
 package handler
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io"
@@ -13,9 +14,9 @@ import (
 var datadogClientFunc = client.NewDatadogClientWithTenancyAndSite
 
 func MyHandler(ctx context.Context, in io.Reader, out io.Writer) {
-	events, err := formatter.Decode(in)
+	raw, err := io.ReadAll(in)
 	if err != nil {
-		log.Printf("Error decoding event batch: %v", err)
+		log.Printf("Error reading event batch: %v", err)
 		writeResponse(out, "error", "", err)
 		return
 	}
@@ -26,14 +27,32 @@ func MyHandler(ctx context.Context, in io.Reader, out io.Writer) {
 		writeResponse(out, "error", "", err)
 		return
 	}
+	url := fmt.Sprintf("https://cloudplatform-intake.%s/api/v2/cloudchanges", site)
+	ociHeaders := map[string]string{"Dd-Oci-Tenancy-Id": tenancyOCID}
+
+	// Backfill mode: drain this region's bucket instead of forwarding new events.
+	if client.IsBackfillTrigger(raw) {
+		if err := ddclient.Backfill(ctx, url, ociHeaders); err != nil {
+			log.Println(err)
+			writeResponse(out, "error", "", err)
+			return
+		}
+		writeResponse(out, "success", "Backfill complete", nil)
+		return
+	}
+
+	events, err := formatter.Decode(bytes.NewReader(raw))
+	if err != nil {
+		log.Printf("Error decoding event batch: %v", err)
+		writeResponse(out, "error", "", err)
+		return
+	}
 
 	payloads, dropped := formatter.Chunk(events)
 	if dropped > 0 {
 		log.Printf("Dropped %d event(s) exceeding the %d byte intake limit", dropped, formatter.MaxBodyBytes)
 	}
 
-	url := fmt.Sprintf("https://cloudplatform-intake.%s/api/v2/cloudchanges", site)
-	ociHeaders := map[string]string{"Dd-Oci-Tenancy-Id": tenancyOCID}
 	for _, payload := range payloads {
 		if err := ddclient.SendMessageToDatadog(ctx, payload, url, ociHeaders); err != nil {
 			log.Printf("Error sending events batch to Datadog: %v", err)
--- a/datadog-functions/events-forwarder/internal/handler/handler.go
+++ b/datadog-functions/events-forwarder/internal/handler/handler.go
@@ -32,12 +32,13 @@ func MyHandler(ctx context.Context, in io.Reader, out io.Writer) {
 
 	// Backfill mode: drain this region's bucket instead of forwarding new events.
 	if client.IsBackfillTrigger(raw) {
-		if err := ddclient.Backfill(ctx, url, ociHeaders); err != nil {
+		summary, err := ddclient.Backfill(ctx, url, ociHeaders)
+		if err != nil {
 			log.Println(err)
-			writeResponse(out, "error", "", err)
+			writeResponse(out, "error", summary.String(), err)
 			return
 		}
-		writeResponse(out, "success", "Backfill complete", nil)
+		writeResponse(out, "success", "Backfill complete: "+summary.String(), nil)
 		return
 	}
 
--- a/datadog-functions/lib/client/client.go
+++ b/datadog-functions/lib/client/client.go
@@ -302,25 +302,42 @@ func IsBackfillTrigger(raw []byte) bool {
 	return json.Unmarshal(raw, &t) == nil && t.BackfillMode == "true"
 }
 
+// BackfillSummary reports what a backfill run accomplished, so the outcome is
+// visible in the function's response and not only in logs.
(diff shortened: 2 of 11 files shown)


datadog-functions/events-forwarder/internal/handler/handler.go
```diff
@@ -26,14 +27,32 @@ func MyHandler(ctx context.Context, in io.Reader, out io.Writer) {
 		writeResponse(out, "error", "", err)
 		return
 	}
+	url := fmt.Sprintf("https://cloudplatform-intake.%s/api/v2/cloudchanges", site)
+	ociHeaders := map[string]string{"Dd-Oci-Tenancy-Id": tenancyOCID}
+
+	// Backfill mode: drain this region's bucket instead of forwarding new events.
+	if client.IsBackfillTrigger(raw) {
+		if err := ddclient.Backfill(ctx, url, ociHeaders); err != nil {
+			log.Println(err)
```
bengillmandd, line 37: Would we want to wrap this err from ddclient.Backfill? Or are the errors all pre-wrapped (ie. have enough context for us to troubleshoot things later on)

datadog-functions/lib/client/client.go
```diff
@@ -282,3 +285,126 @@ func putObject(ctx context.Context, osClient objectStorageAPI, namespace, bucket
...
+
+			if err := client.replayWithRetry(ctx, data, intakeURL, extraHeaders...); err != nil {
+				// Unrecoverable after one retry. Leave the object in place and stop
+				// this run; hubmanager will invoke again shortly.
+				return fmt.Errorf("backfill: replay failed for object %q, stopping run: %w", name, err)
+			}
+
+			if err := deleteObject(ctx, osClient, namespace, bucket, name); err != nil {
+				// Delivered but not deleted: it will be re-sent next run (at-least-once).
+				log.Printf("backfill: delivered but failed to delete object %q: %v", name, err)
```
bengillmandd: I'm curious why we would print this instead of creating a list of errors and sending them upon completion?

datadog-functions/lib/client/client.go
```diff
@@ -282,3 +285,126 @@ func putObject(ctx context.Context, osClient objectStorageAPI, namespace, bucket
...
+
+		for _, summary := range resp.Objects {
+			if summary.Name == nil {
+				continue
+			}
+			name := *summary.Name
+
+			data, err := getObject(ctx, osClient, namespace, bucket, name)
+			if err != nil {
+				log.Printf("backfill: failed to read object %q, skipping: %v", name, err)
```
bengillmandd: same as below, why print instead of create a list of errors?

datadog-functions/lib/client/client.go
```diff
@@ -33,36 +34,42 @@ type DatadogClient struct {
...
+	}
+	return err
+}
+
+// replaySend sends a payload read back from a backfill bucket. Unlike
+// SendMessageToDatadog it never re-persists on a 5xx (the payload is already in
+// the bucket), so a failed replay just returns its status and error and is left
+// in place for the next run. The HTTP status lets the caller distinguish
+// throttling (429) from other failures.
+func (client *DatadogClient) replaySend(ctx context.Context, message []byte, url string, extraHeaders ...map[string]string) (int, error) {
```
bengillmandd, line 59: Why wrap send like this?

bengillmandd (approved):

datadog-functions/events-forwarder/internal/handler/handler.go
erics-ddog: They're pre-wrapped! See `func Backfill` in `client.go`

datadog-functions/lib/client/client.go
erics-ddog: I wanted to keep the forward-vs-replay send distinction clear, especially if we ever decide to give `replaySend` its own extra functionality. I can inline it if you feel strongly about it- no functional difference for now.

datadog-functions/lib/client/client.go
erics-ddog: Good call on visibility. Kept the log-and-continue control flow so that we don't abort on a single unreadable/undeletable object, but Backfill now returns a `BackfillSummary{replayed, skipped, delete_failures}` for better visibility.

datadog-functions/lib/client/client.go
EamonBrady1: This isn't a Datadog internal server error, we shouldn't treat this as 5xx

EamonBrady1: same with this

datadog-functions/lib/client/client.go
```diff
@@ -92,13 +100,6 @@ func (client *DatadogClient) sendMessage(ctx context.Context, message []byte, ur
 		if err == nil {
 			log.Printf("Error response body: %s", string(body))
 		}
-		// On a 5xx the payload is undeliverable right now. Persist it to the
-		// per-data-type backfill bucket if one exists so it can be replayed
-		// later; otherwise drop it for now. This is best-effort and does not
-		// change the status/error returned below.
-		if resp.StatusCode >= 500 && resp.StatusCode < 600 {
-			handleServerErrorPayload(ctx, message, url)
```
EamonBrady1, line 100: we should be checking for a 5xx error here, since this is the status code returned from the call to Datadog. Anything else is OCI processing prior to reaching Datadog

PR #150, opened 2026-06-29
add event_collection_enabled variable to datadog-integration and datadog-terraform-onboarding

yuhuyoyo:
## Summary

- Adds `event_collection_enabled` (bool, default `false`) to both `datadog-integration` and `datadog-terraform-onboarding` stacks
- Variable passes through the module chain: root `variables.tf` → `main.tf` → `modules/integration/variables.tf` → `modules/integration/locals.tf`
- Sets `events_collection_enabled` in the Datadog REST API payload, enabling OCI Events Service collection at onboarding time
- Added to `datadog-integration/schema.yaml` so it appears in the OCI Resource Manager UI
- Default is `false` — no behavior change for existing deployments

## Why

Supports a new UI toggle on the OCI onboarding form ("Configure Events Collection" step) that lets users opt into resource change collection during initial setup. Datadog is the source of truth for this value after creation; it is not read back into UI state.

## Test plan

- [x] `terraform validate` passes in both `datadog-terraform-onboarding/` and `datadog-integration/`
- [x] Apply `datadog-terraform-onboarding` with `event_collection_enabled = true` and verify `events_collection_enabled: true` appears in the tenancy payload sent to Datadog
- [ ] Apply with default (`false`) and verify existing behavior is unchanged
- [x] Confirm `schema.yaml` renders the new variable in ORM stack UI


datadog-integration/modules/integration/locals.tf
```diff
@@ -17,13 +17,15 @@ locals {
...
     }
+    events_collection_enabled : var.events_collection_enabled
     defined_tags : [for k, v in var.defined_tags : "${k}:${v}"]
   }
   logs_only_attributes = { for k, v in {
     resource_collection_enabled : false
     metrics_config : {
       enabled : false
     }
+    events_collection_enabled : false
```
EamonBrady1, line 28: Curious what Ellie decides on the route for this one.

datadog-integration/schema.yaml
```diff
@@ -117,11 +118,19 @@ variables:
     required: false
     default: false
 
+  events_collection_enabled:
+    title: Event Collection Enabled
+    type: string
+    description: |
+      Indicates if event collection (OCI Events Service) should be enabled. Note: this setting is ignored when Logs Only is enabled.
```
EamonBrady1, line 125: I think this is a good way to handle it, makes it clear

datadog-terraform-onboarding/modules/integration/locals.tf
```diff
@@ -15,6 +15,7 @@ locals {
     logs_config : {
       Enabled = var.logs_enabled
     }
+    events_collection_enabled : var.events_collection_enabled
     defined_tags : [for k, v in var.defined_tags : "${k}:${v}"]
   }
   logs_only_attributes = { for k, v in {
```
EamonBrady1, line 21: I think we need to add `events_collection_enabled : false` here

yuhuyoyo: yes good catch!

EamonBrady1 (approved):
We'll also have to add the update handler here in OIA: https://github.com/ddoghq/dd-source/blob/main/domains/cloud_platform/oci/apps/apis/oci-integrations-api/internal/tenancyconfig/update_tenancy.go#L76-L80

datadog-integration/modules/integration/locals.tf
yuhuyoyo: based on the slack thread comment, i believe we are on the same page that log only exclude events as well.

EamonBrady1: Yep that sounds good to me

PR #158, opened 2026-07-22
ECI-1919: per-region Vault with home-region fallback

yuhuyoyo:
## Summary
- Vault/secret access in OCI is region-scoped, so a region's function couldn't read the home region's vault secret. Each region now creates its own independent Vault + Key + Secret when it has spare quota, falling back to the home-region vault/secret otherwise — implemented in both `datadog-integration` (Quickstart/ORM) and `datadog-terraform-onboarding` (raw Terraform module).
- Per-region Vault creation is opt-in via `enable_regional_vaults` (default `false`); parent stack waits for all regional stacks to finish before completing its own apply.
- Fixes a self-consuming quota gate: since a region's own vault consumes one unit of the very quota `create_regional_vault` checks, a region with exactly one spare vault would create it, then flip back to `false` on the next apply and destroy/recreate it — flapping indefinitely. The decision is now sticky once the vault exists in state.
- Fixes `datadog-integration/delete_stack.sh`: it only checked whether `create-destroy-job` returned a job ID, not the job's actual terminal `lifecycle-state`, so a failed destroy job still deleted the ORM stack record, orphaning undestroyed resources. It now checks the job's lifecycle-state and leaves the stack in place on failure.
- Collapses `datadog-terraform-onboarding/vault_quota.tf` from 813 lines of per-region static blocks to 52 lines using `for_each` over `data "external"` sources (no per-region Terraform provider alias needed for these, unlike the regional deployment modules).
- Fixes `datadog-integration/regional_stack.tf`: ORM stack `--variables` is `map<string,string>`; passing the unquoted boolean `enable_regional_vaults` caused OCI to silently drop that key from stored stack variables (no error), so the flag never reached regional child stacks. Now quoted as a string.
- Fixes a related gap in `datadog-integration/regional_stack.tf`: when a regional child stack already existed, the code only re-ran `create-apply-job` against its stale config — any variable change (including the fix above) was silently ignored on re-apply. Now calls `stack update` to refresh config/variables before re-applying, with error logging on failure.

## Test plan
- [x] `terraform validate` passes for both `datadog-integration/modules/regional-stacks` and `datadog-terraform-onboarding` (root + nested module)
- [x] `terraform apply` run end-to-end against a real OCI tenancy (raw Terraform onboarding path), confirmed regional vault fallback logic and no duplicate home-region vault
- [x] Re-ran `terraform plan` against the applied state after the sticky-quota-gate fix — confirmed no vault/key/secret resources are planned for destroy/recreate (only unrelated in-place attribute drift)
- [x] Confirmed live in OCI (`eu-frankfurt-1`, `ap-tokyo-1`) that orphaned vault/function resources from a prior destroy bug matched the `delete_stack.sh` root cause
- [x] Quickstart (ORM) path: opted in to `enable_regional_vaults` on a real ORM stack, verified all 9 subscribed regions created a fresh regional Vault/Key/Secret, and confirmed the stack-update-on-reuse fix by observing regional stack variables actually refresh on re-apply
- [x] Quickstart (ORM) path: opted back out (`enable_regional_vaults = false`) and re-applied, verified via decoded apply job logs that all 9 regions destroyed their regional Vault/Key/Secret and each function's `API_KEY_SECRET_OCID` fell back to the home-region secret
- [ ] Exercise the quota-exhaustion fallback path against genuinely exhausted vault quota (not yet tested against real exhausted quota)


datadog-integration/modules/regional-stacks/locals.tf
```diff
@@ -10,10 +10,26 @@ locals {
...
+  # The region/secret actually backing this function's Datadog API key: this
+  # region's own vault when created, otherwise the home-region vault.
+  vault_region      = local.create_regional_vault ? var.region : var.home_region
+  api_key_secret_id = local.create_regional_vault ? oci_vault_secret.api_key[0].id : var.api_key_secret_id
+
   config = {
     "DD_SITE"                  = var.datadog_site,
-    "HOME_REGION"              = var.home_region,
-    "API_KEY_SECRET_OCID"      = var.api_key_secret_id
+    "HOME_REGION"              = local.vault_region,
```
EamonBrady1, line 31: Are we using this home region variable for anything except the vault? I think Eric's backfill code determines regions automatically, but definitely should confirm this won't break any function logic

datadog-integration/modules/regional-stacks/locals.tf
```diff
@@ -10,10 +10,26 @@ locals {
...
+    "API_KEY_SECRET_OCID"      = local.api_key_secret_id
```
EamonBrady1, line 32: Also I assume you did, but did we test the fallback to using the home region vault?

yuhuyoyo: i tested the false scenario which all back to home region. now i'm creating fake vault token to test the fallback scenario

datadog-integration/regional_stack.tf
```diff
@@ -44,28 +44,21 @@ resource "null_resource" "regional_stacks_create_apply" {
       exit 0
     fi
 
-    # Do not wait for the stack to be applied except for home region
-    WAIT_COMMAND=""
-    if [[ "$CURRENT_REGION" == "${local.home_region_name}" ]]; then
-      WAIT_COMMAND="--wait-for-state SUCCEEDED --wait-for-state FAILED"
-    fi
```
EamonBrady1, line 51: So does deleting this mean that we _do_ wait now?

yuhuyoyo: yes, we need to make sure the vaults deletion (which observes to be flaky) are successful in the regional vault. Especially we clean up the regional stack afterward, if we don't wait, it will appears to be clean up but left litters around that we don't want.

datadog-integration/variables.tf
```diff
@@ -106,3 +106,9 @@ variable "defined_tags" {
   description = "Defined tags to apply to all created resources. One entry per line in the format namespace.key:value (e.g. CostCenter.Environment:prod). Leave blank unless your tenancy has mandatory tag defaults."
   default     = ""
 }
+
+variable "enable_regional_vaults" {
```
EamonBrady1, line 110: Seems like we shouldn't allow customers to enable this if they already onboarded right? I think enabling it after originally being on only one vault would just cause all kinds of issues with function references and which API keys the functions try to read from

EamonBrady1: For example if we create a function for OCI events, is hubmanager properly pointing it to the correct regional vault? Does hubmanager have any way of knowing that an account uses regional vaults?

datadog-integration/modules/regional-stacks/locals.tf
yuhuyoyo: tested

datadog-integration/variables.tf
yuhuyoyo: for existing customer, new vault will be created and functions are updated in place. good callout for events. because the env variable are inherited from the application, no treatment is needed in hubmanager for events.

datadog-integration/modules/regional-stacks/locals.tf
yuhuyoyo: vault is the only use case in datadog-function https://github.com/DataDog/oracle-cloud-integration/blob/9e547ecca1eb173300ff3e4f99dbf368aec88fe5/datadog-functions/lib/client/client.go#L129

datadog-integration/variables.tf
EamonBrady1: Gotcha, so existing logs/metrics functions will re-route to the new vaults?

For events, if we create an events function for an existing customer, then they reapply the stack to set up multiple vaults, wouldn't the events config still point to the old vault? I understand that it pulls the configuration from the application, but updating the application wouldn't automatically update every function inside of it, right?

yuhuyoyo: the variables are inherited from the application.

EamonBrady1 (approved):

PR #163, opened 2026-08-10
move the multi-region vault option out of advance config

yuhuyoyo:
**What**: move the multi-region vault selection out of the advanced config

Screenshot 2026-08-10 at 1 45 56 PM

**Why**: when we pass the enable_regional_vault variable to create new oci stack, it get ignored if advanced config is not enabled but it's weird we by-default enable advance config.


EamonBrady1 (approved):
LGTM, does this need a version bump though? I'm good with it either way though

PR #164, opened 2026-08-13
Emit forwarder version on request (hubmanager version-check)

erics-ddog:
## What
- Each forwarder recognizes a new control message `{"version_check":"true"}` (mirroring the existing `{"backfill_mode":"true"}` pattern) and responds with its build-stamped `Version` directly in the invoke response — no Vault fetch, no HTTP call to Datadog, nothing that touches the customer's Datadog org.
- `Version` is still stamped into the binary at build time via `-ldflags "-X main.Version=<tag>"` (`Dockerfile-events`, `Dockerfile-logs`, `Dockerfile-metrics`, `docker-images.sh`), so it reflects what's actually baked into the image, not what OCI's function record claims should be running.
- `fnResponse` gains an optional `version` field; a new `writeVersionResponse` answers the check.
- README's manual build instructions now pass `--build-arg TAG=<tag>` so a locally built image doesn't silently report `version:unknown`.

## Why (redesign from the original approach)
This PR originally had each forwarder push a `datadog.oci.forwarder.cold_start` metric to Datadog's intake on cold start, using the same client/credentials the forwarder already uses to send the customer's own telemetry. Review feedback (see thread below) correctly caught two real problems with that:
1. It would submit a new custom metric under the *customer's* Datadog org/API key — billable to them, and invisible to Datadog engineers in aggregate unless someone went looking inside a specific customer's org.
2. It ran synchronously before `fdk.Handle`, chained through Vault's 3-minute timeout, so a slow/unavailable Vault or Datadog could delay or fail a cold start over a best-effort observability signal.
3. (Separately) it tagged region from `HOME_REGION`, which is the Vault's home region, not necessarily the function's actual execution region in multi-region deployments — collapsing the very rollout signal the metric existed to produce.

This redesign flips the direction: instead of the forwarder pushing to Datadog, hubmanager — which already talks to these functions over OCI's control/invoke plane using Datadog's own credentials, and already does exactly this for backfill triggers — pulls the version by invoking the function synchronously and reading the response. See the companion PR in `dd-source`: https://github.com/ddoghq/dd-source/pull/58292

This resolves all three points above as a side effect: nothing here ever touches customer credentials or the customer's Datadog org (point 1), there is no network/Vault call anywhere on this path so nothing can block `fdk.Handle` (point 2), and region tagging moves entirely to hubmanager's side, using its own known-correct region rather than a self-reported env var (point 3).

## Testing
- [x] `go build ./...` / `go vet ./...` clean for all three forwarders and `lib`.
- [x] `go test ./...` passes for all three forwarders and `lib`, including new `TestMyHandler_VersionCheck` tests per forwarder that assert the check succeeds with **no** environment variables set at all (proving it can't depend on Vault/Datadog config).
- [x] `gofmt -l` clean on all changed files.
- [ ] Not yet verified against a live OCI region — this is still a draft pending the companion hubmanager PR.

## Notes
- Companion PR (hubmanager side: invokes `{"version_check":"true"}` synchronously, emits the desired-vs-actual comparison as an internal `dd.oci.hubmanager.*` metric): https://github.com/ddoghq/dd-source/pull/58292


EamonBrady1 (changes requested):
This would be a pretty notable change and is sending a custom metric to Datadog from the customer account (which would charge them). We currently don't have any behavior in the forwarding functions that emits monitoring data directly to Datadog.
Additionally, if we are sending it with this client, it would appear as monitoring data in their account, invisible to us unless we partlow into user accounts every time we want to check this, which we wouldn't want to do.
I think the solution to this issue is a bit more complicated than just emitting a metric from the forwarding functions and is part of bigger picture OCI monitoring that we've thought about in the past. It might require a new endpoint to send this monitoring data to or something like that, but I think we need to think about this one a little more

erics-ddog:
@EamonBrady1 you were right — I've pushed a redesign that removes the push-to-Datadog-intake entirely. Forwarders no longer touch Vault or send anything to the customer's Datadog org for this. Instead they just echo their build-stamped version back when hubmanager invokes them with `{"version_check":"true"}` (same pattern as the existing backfill trigger), and hubmanager itself — using its own OCI credentials — is the one that emits the comparison as an internal metric. Companion PR in `dd-source` coming shortly.

@bengillmandd this also addresses your synchronous-call concern from a different angle than "make it async": there's no longer any Vault/HTTP call on the cold-start path at all, so nothing can block `fdk.Handle`.

Also fixed the two Codex findings: region is no longer self-reported via `HOME_REGION` (that move to hubmanager's side entirely), and the README's manual build instructions now pass `--build-arg TAG`.

Keeping this as a draft until the hubmanager-side PR is up so they can be reviewed together.

erics-ddog:
Companion PR is up: https://github.com/ddoghq/dd-source/pull/58292 (hubmanager side — invokes `{"version_check":"true"}` synchronously via the OCI invoke plane and emits the desired-vs-actual comparison as an internal `dd.oci.hubmanager.function_version_check` metric). Both are ready for review together.

datadog-functions/docker-images.sh
```diff
@@ -60,6 +60,7 @@ build_and_push_image() {
   echo "Building and pushing multi-arch image ${IMAGE_PATH} to $REGISTRY..."
   docker buildx build -f ${DOCKER_FILE} \
     --platform linux/amd64,linux/arm64 \
+    --build-arg TAG="${TAG}" \
```
EamonBrady1: What is this doing?

datadog-functions/events-forwarder/internal/handler/response.go
```diff
@@ -40,3 +41,19 @@ func writeResponse(out io.Writer, status string, message string, err error) {
...
+func writeVersionResponse(out io.Writer, version string) {
+	fdk.AddHeader(out, "Content-Type", "application/json")
+	jsonResp, err := json.Marshal(fnResponse{Status: "success", Version: version})
+	if err != nil {
+		log.Printf("Error marshalling version response: %v", err)
+		fdk.WriteStatus(out, 500)
+		out.Write([]byte(`{"status":"error","error":"Internal server error"}`))
+		return
+	}
+	fdk.WriteStatus(out, 200)
```
EamonBrady1: do these statuses actually make it to hubmanager? The idea seems right, but I'm curious if this is actually what's surfaced as the function response when we invoke it

PR #165, opened 2026-08-14
Add OCI cleanup foundation

johannzhang8:
## Summary
- add a safe CLI facade for deleting resources when missing Terraform state
- add ownership discovery, resumable manifests, dry-run execution gates, and actionable OCI errors
- keep cleanup tooling in root-level `oci-integration-cleanup/` so it is excluded from the integration release zip
- For context on how this package is intended to work - [README](https://github.com/DataDog/oracle-cloud-integration/blob/3d524de2a23d8be50a2d967e05a7c5992699e1c5/oci-integration-cleanup/README.md)

No need to review extras.py, that was reviewed in #171 

## Test plan
- [x] `python3 -m unittest discover -s oci-integration-cleanup/tests -p 'test_*.py' -v` (13 tests)


johannzhang8 (commented):
test review comment - will need cleanup

johannzhang8 (commented):
## Review notes

Note: my token can't create inline/diff comments in this EMU-managed org (both the REST `pulls/comments` endpoint and the GraphQL `addPullRequestReview(threads:...)` mutation return `403 Unauthorized: As an Enterprise Managed User, you cannot access this content`), so posting as a single review comment instead, organized by file.

### Correctness

**`oci_cleanup/oci.py:56`** — `"404" in stderr` is a substring match against raw stderr text. Any unrelated OCI error whose message happens to contain the digits "404" (e.g. inside an OCID, request ID, or error-body dump) would be silently treated as "resource already deleted" and swallowed instead of surfacing as a real failure. Consider matching on a parsed OCI error code/status field rather than a free-text substring search.

**`oci_cleanup/oci.py:51`** — `json.loads(output) if output else {}` assumes successful OCI CLI output is always valid JSON. If a subcommand ever returns non-JSON (or malformed) output on a zero exit code, this raises a raw `JSONDecodeError` that bypasses `CommandError`, losing the service/region/customer-action diagnostics `action()` otherwise records for failures.

**`oci_cleanup/discovery.py:167`** (in the compartment-resolution block starting ~128) — when `compartment_candidates` is empty and `--compartment-id` is supplied explicitly, this accepts `explicit_signal` with no cross-check at all; the "conflicts with tagged resource compartments" guard a few lines above only fires when candidates is non-empty. A typo'd `--compartment-id` with zero discoverable tagged resources in it would be silently accepted as the cleanup target, bypassing the tag-based ownership safety net entirely.

**README vs. implementation** — the README documents modules/behavior that don't exist yet in this PR (`domains/extras.py`, `domains/network.py`, `domains/kms.py`, the nested-blocker discovery flow), and `region.py`'s `cleanup_region` is currently a no-op stub. Understandable for a stacked PR, but worth a note that the README describes the end state of the whole stack, not this PR alone, so reviewers don't cross-check it against PR #165 in isolation.

**`tests/test_integration_cleanup.py::test_run_can_clean_regions_concurrently`** — monkeypatches methods (`discover_extra_resources`, `_ask_yes_no`, `cleanup_home_identity`, `cleanup_tags`, `cleanup_parent_stack`, `delete_compartment`) that don't exist on the class yet, and asserts `_ask_yes_no` is never called. Since `engine.run()` never invokes any of them in this PR, the assertion is vacuously true today — the "no interactive prompts from parallel region workers" property isn't actually tested until later PRs wire these calls in.

### Style / simplification

**`oci_cleanup/cli.py:81`** — dead `elif args.delete_compartment: pass` branch does nothing. Either drop it (keeping the comment if useful) or wire up the described behavior.

**`oci_cleanup/base.py:50`** (`_read_action`) — no call site in this PR; looks like dead code shipped ahead of a consumer.

**`oci_cleanup/discovery.py:134`** — `resource.get("resource-type") or resource.get("resource_type") or ""` repeated inline here and again near line 239. Consider promoting to a small helper alongside the existing `resource_name`/`resource_compartment` accessors in `resources.py`.

**`integration_cleanup.py` / `oci_cleanup/__init__.py`** — both hand-maintain ~50-line duplicate `__all__` lists; risk of silent drift as more symbols are added across the stack.

**`oci_cleanup/manifest.py` + `base.py::action()`** — `record_action(...)` and `manifest.save()` acquire the manifest lock as two separate critical sections rather than one atomic read-modify-write. Harmless today, but worth tightening if concurrent-region logic grows.

oci-integration-cleanup/oci_cleanup/resources.py
```diff
@@ -0,0 +1,159 @@
...
+        or resource.get("secret-name")
+        or resource.get("secret_name")
+        or resource.get("name")
+        or ""
+    )
+
+
+def resource_compartment(resource: dict[str, Any]) -> str:
+    return str(
+        resource.get("compartment-id")
```
yuhuyoyo: this is crazy how inconsistent the field name is!

oci-integration-cleanup/oci_cleanup/resources.py
```diff
@@ -0,0 +1,159 @@
...
+        or resource.get("state")
+        or ""
+    ).upper()
+
+
+def is_deleted_or_deleting(resource: dict[str, Any]) -> bool:
+    return lifecycle_state(resource) in {
+        "DELETED",
+        "DELETING",
+        "DETACHED",
```
yuhuyoyo: detach is not a deleting state

oci-integration-cleanup/README.md
```diff
@@ -0,0 +1,203 @@
...
+
+The tool removes OCI resources only. Disable or remove the OCI integration in
+Datadog before running execute mode; otherwise the Datadog control plane can
+recreate managed connectors, buckets, event rules, or streams.
+
+Development is split across a stacked PR series. This README describes the
+completed tool; intermediate PRs contain only the modules enabled at that layer
+of the stack.
+
+## Files
```
yuhuyoyo: can remove this section. the source of truth should be the code in this folder.

oci-integration-cleanup/integration_cleanup.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+def parse_args(argv: Optional[list[str]] = None) -> argparse.Namespace:
+    parser = argparse.ArgumentParser(
+        description=(
+            "Discover and safely clean a Datadog OCI Quickstart installation "
+            "when Terraform state is unavailable."
+        )
+    )
+    parser.add_argument("--tenancy-id", required=True)
```
yuhuyoyo: the profile that user specified should be pointing to a tenancy already so we wouldn't need to specify tenancy-id for the oci calls.

oci-integration-cleanup/integration_cleanup.py
```diff
@@ -0,0 +1,126 @@
...
+    parser.add_argument("--compartment-id")
+    parser.add_argument("--domain-endpoint")
+    parser.add_argument("--parent-stack-id")
+    parser.add_argument("--profile")
+    parser.add_argument("--oci-bin", default=os.getenv("OCI_BIN", "oci"))
```
yuhuyoyo: or shutil.which("oci")

oci-integration-cleanup/integration_cleanup.py
```diff
@@ -0,0 +1,126 @@
...
+def parse_args(argv: Optional[list[str]] = None) -> argparse.Namespace:
+    parser = argparse.ArgumentParser(
+        description=(
+            "Discover and safely clean a Datadog OCI Quickstart installation "
+            "when Terraform state is unavailable."
+        )
+    )
+    parser.add_argument("--tenancy-id", required=True)
+    parser.add_argument("--compartment-id")
+    parser.add_argument("--domain-endpoint")
```
yuhuyoyo: what is domain-endpoint? some help text would be nice

oci-integration-cleanup/integration_cleanup.py
```diff
@@ -0,0 +1,126 @@
...
+
+def parse_args(argv: Optional[list[str]] = None) -> argparse.Namespace:
+    parser = argparse.ArgumentParser(
+        description=(
+            "Discover and safely clean a Datadog OCI Quickstart installation "
+            "when Terraform state is unavailable."
+        )
+    )
+    parser.add_argument("--tenancy-id", required=True)
+    parser.add_argument("--compartment-id")
```
yuhuyoyo: is this OCID or the compartment name? some help text on this is the root compartment where the Stack is installed on would be nice. But also if we have the stack, can we figure out the compartment for the user?

oci-integration-cleanup/integration_cleanup.py
```diff
@@ -0,0 +1,126 @@
...
+    EngineMixin,
+    DiscoveryMixin,
+    ExtrasMixin,
+    RegionMixin,
+    CleanupBase,
+):
+    """Discover unexpected resources in one OCI Quickstart installation."""
+
+
+def parse_args(argv: Optional[list[str]] = None) -> argparse.Namespace:
```
yuhuyoyo: a skill for customer to scan the tenancy and figure out the OCIDs to pass in would be nice.

oci-integration-cleanup/integration_cleanup.py
```diff
@@ -0,0 +1,126 @@
...
+            "Use a small value such as 4 to limit OCI throttling."
+        ),
+    )
+    parser.add_argument(
+        "--state-file",
+        type=pathlib.Path,
+        help="Execution manifest path (required for --execute)",
+    )
+    parser.add_argument(
+        "--execute",
```
yuhuyoyo: can we name this --dry-run. it's more obvious.

oci-integration-cleanup/README.md
```diff
@@ -0,0 +1,203 @@
...
+
+Every mutation passes through `oci_cleanup/base.py`. Successful actions are
+recorded in `oci_cleanup/manifest.py` and skipped on later runs unless OCI still
+lists the resource. Failed and blocked actions remain retryable.
+
+Customer-facing output contains a concise service and operation summary. The
+state file retains the complete OCI command output and request ID under
+`raw_error` for troubleshooting.
+
+## Usage
```
yuhuyoyo: i think this readme should just keep this section. it's the most important for a customer to read.

oci-integration-cleanup/oci_cleanup/manifest.py
```diff
@@ -0,0 +1,107 @@
+"""Responsibility: load and persist resumable cleanup state.
+
+Safety boundary: rejects tenancy mismatches and serializes updates under a lock.
+Cleanup sequence role: records decisions and actions across the full cleanup sequence.
+
+``Manifest.load`` initializes or validates tenancy-bound JSON state. Thread-safe
```
yuhuyoyo: why do we need a manifest file? for the lock argument, there's one thread per region so I don't think we need a lock. for resumable perspective, instead of manifest, a simpler flow will just be delete -> 404 -> treat as success -> continue.

oci-integration-cleanup/oci_cleanup/domains/extras.py
```diff
@@ -0,0 +1,747 @@
...
+delete approved children or preserve their parents.
+"""
+
+from __future__ import annotations
+
+import sys
+from dataclasses import dataclass
+from typing import Any, Optional
+
+from ..constants import (
```
yuhuyoyo: if there's a new resourecs that may have a parent/child hiearchy or a new function added, we will need to remember to update this script. It will increase maintenance burdon. Can we instead just catch the error in OCI when trying to delete these resources. for example, when deleting a function app that still have functions in it, we will fail and just notify the user that this function apps need manual clean up.

johannzhang8: hmm this is true. I guess in the case of functions, the change would be adding the name of the function to function_names. Additionally, if there is a unknown function, the user will be prompted if they want to delete that function. For a new resource, I know we added some buckets recently, so I think we could just add that bucket name to the BACKFILL_BUCKET_NAMES (probably needs a rename). I will update the script to explicitly tell the user to handle deletion if the script fails to delete any object, which should be sufficient for when we add new resources.

oci-integration-cleanup/README.md
johannzhang8: yes, the other stuff was for reviewers

oci-integration-cleanup/oci_cleanup/resources.py
johannzhang8: If a customer has added extra vnics, we want to detach instead of delete them, so as far as the script cares, detached is the same as deleted.

oci-integration-cleanup/integration_cleanup.py
johannzhang8: compartment OCID. Will make this more clear. For the stack thing, I was under the assumption that this has to work in the case we cannot rely on the stack at all.

oci-integration-cleanup/integration_cleanup.py
johannzhang8: It is the identity domain endpoint. Optional parameter. Will update the usage doc

oci-integration-cleanup/oci_cleanup/resources.py
johannzhang8: ah this is mainly due to python normalization. OCI json should return compartment-id but just to be safe i include compartment_id

oci-integration-cleanup/integration_cleanup.py
johannzhang8: Additionally, we seem to be pretty good at getting domain endpoint. Will remove

oci-integration-cleanup/integration_cleanup.py
johannzhang8: will do that in separate PR

oci-integration-cleanup/oci_cleanup/resources.py
yuhuyoyo: i see. thanks for explainng

oci-integration-cleanup/README.md
yuhuyoyo: then can you clean it up once you are done with your chains of PRs?

yuhuyoyo (approved):

oci-integration-cleanup/README.md
johannzhang8: yes it is cleaned now. README in the description will still work though for reference (it is tied to commit before).

PR #166, opened 2026-08-14
Clean up Quickstart networking

johannzhang8:
## Summary
- delete owned subnets, route rules and tables, gateways, and VCNs in dependency order
- discover nested VNIC and network blockers and require explicit confirmation before destructive cleanup
- preserve customer-owned resources and stop parent deletion when dependencies remain

## Stack
- Depends on https://github.com/DataDog/oracle-cloud-integration/pull/165

## Test plan
- [x] `python3 -m unittest discover -s oci-integration-cleanup/tests -p 'test_*.py' -v` (26 tests)

### DIFF
--- a/oci-integration-cleanup/integration_cleanup.py
+++ b/oci-integration-cleanup/integration_cleanup.py
@@ -26,6 +26,7 @@ from oci_cleanup import __all__ as package_api
 from oci_cleanup.base import CleanupBase
 from oci_cleanup.discovery import DiscoveryMixin
 from oci_cleanup.domains.extras import ExtrasMixin
+from oci_cleanup.domains.network import NetworkMixin
 from oci_cleanup.engine import EngineMixin
 from oci_cleanup.region import RegionMixin
 
@@ -37,9 +38,10 @@ class QuickstartCleanup(
     DiscoveryMixin,
     ExtrasMixin,
     RegionMixin,
+    NetworkMixin,
     CleanupBase,
 ):
-    """Discover unexpected resources in one OCI Quickstart installation."""
+    """Discover and remove Quickstart networking resources."""
 
 
 def _parse_bool(value: str) -> bool:
--- a/oci-integration-cleanup/oci_cleanup/domains/network.py
+++ b/oci-integration-cleanup/oci_cleanup/domains/network.py
@@ -1,0 +1,388 @@
+"""Responsibility: clean the Datadog VCN, subnet, gateways, and route dependencies.
+
+Safety boundary: validates exact ownership and blocks parents with unapproved nested resources.
+Cleanup sequence role: runs after forwarding services and before regional KMS cleanup.
+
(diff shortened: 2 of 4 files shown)


johannzhang8 (commented):
## Review notes

(Posting as a single review comment, not inline — this org's EMU/SSO policy blocks the GitHub API's diff-anchored/inline comment endpoints for my token: `403 Unauthorized: As an Enterprise Managed User, you cannot access this content` on both the REST `pulls/comments` endpoint and the GraphQL `addPullRequestReview(threads:...)` mutation.)

### Correctness

**`oci_cleanup/domains/extras.py:96`** (`_compute_compartment_ids`) — if the `iam compartment list` call raises `CommandError` (transient network blip, rate limit), the `except` here just logs a warning and falls through with `compartments = []`. The resulting reduced ID list (`[compartment_id, tenancy_id]`) is memoized into `self._accessible_compartment_ids` (line 109) for the rest of the run — a transient failure gets permanently cached as "these are the only accessible compartments." Later cross-compartment VNIC-attachment lookups silently lose coverage for the rest of the run, which could misclassify a customer compute instance's VNIC as `unsupported-vnic` instead of `compute-instance`. Recommend not caching on error (or failing the run loudly instead).

**`oci_cleanup/domains/network.py:267`** — non-default route-table deletion uses `is_owned(route_table)` alone, while every other resource type in this file (VCN, subnet, gateways) is filtered with `exact_owned(...)` (name + compartment match, not just a freeform tag). This is an inconsistent safety boundary — worth confirming `is_owned` is at least as strict, or unifying on `exact_owned` here too.

**`oci_cleanup/domains/network.py:70`** (`network_dependencies_blocked`) — one unapproved "extra" resource anywhere in a region's network graph blocks teardown of the *entire region's* network cleanup, not just its own subtree (see the `any(...)` over `self.extra_candidates` and the early-return path when `approved_extras_deleted` is false). This looks like an intentional fail-safe, but it's worth confirming that's the intended blast radius — a single unrelated blocker elsewhere in the region currently halts unrelated resource teardown too.

### Style / simplification

Repeated hyphen/underscore field-normalization pattern (`x.get("a-b") or x.get("a_b") or ""`) appears 6+ times across `network.py` and `extras.py` (default-route-table-id, vcn-id, subnet-id, instance-id, etc). Consider a single `field(resource, name)` helper to reduce risk of a copy-paste typo silently breaking one of the fallbacks.

Gateways are listed via separate OCI CLI calls once during discovery (`_discover_network_extras`) and again during deletion (`cleanup_network`). Not a bug, but if `context` already caches per-region resource lists, this could be deduplicated to cut OCI API calls roughly in half for gateways/VCNs/subnets.

The `gateway_specs` tuple in `_discover_network_extras` overloads `expected_name=None` to mean "always treat as extra" vs. a real name meaning "check ownership first" — a named field (e.g. `always_flag: bool`) or a small dataclass would make the downstream branch easier to read than inferring behavior from falsiness.

EamonBrady1 (approved):

PR #167, opened 2026-08-14
Clean up forwarding services and storage

johannzhang8:
## Summary
- remove Datadog service connectors, event rules, streams, Functions resources, and backfill buckets
- verify ownership before deletion and preserve customer applications and buckets
- require approval for unexpected functions that block application deletion

## Stack
- Depends on https://github.com/DataDog/oracle-cloud-integration/pull/166

## Test plan
- [x] `python3 -m unittest discover -s oci-integration-cleanup/tests -p 'test_*.py' -v` (31 tests)

### DIFF
--- a/oci-integration-cleanup/integration_cleanup.py
+++ b/oci-integration-cleanup/integration_cleanup.py
@@ -27,6 +27,7 @@ from oci_cleanup.base import CleanupBase
 from oci_cleanup.discovery import DiscoveryMixin
 from oci_cleanup.domains.extras import ExtrasMixin
 from oci_cleanup.domains.network import NetworkMixin
+from oci_cleanup.domains.services import ServicesMixin
 from oci_cleanup.engine import EngineMixin
 from oci_cleanup.region import RegionMixin
 
@@ -38,10 +39,11 @@ class QuickstartCleanup(
     DiscoveryMixin,
     ExtrasMixin,
     RegionMixin,
+    ServicesMixin,
     NetworkMixin,
     CleanupBase,
 ):
-    """Discover and remove Quickstart networking resources."""
+    """Remove Quickstart forwarding services, storage, and networking."""
 
 
 def _parse_bool(value: str) -> bool:
--- a/oci-integration-cleanup/oci_cleanup/discovery.py
+++ b/oci-integration-cleanup/oci_cleanup/discovery.py
@@ -13,11 +13,13 @@ from __future__ import annotations
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from typing import Any
 
-from .constants import AUTO_COMPARTMENT_NAME, LOGGER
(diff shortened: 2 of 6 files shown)


johannzhang8 (commented):
## Review notes

(Posting as a single review comment, not inline — see note on PR #165/#166: this org's EMU/SSO policy blocks diff-anchored comment endpoints for my token.)

### Correctness

**`oci_cleanup/domains/services.py:289`** (`cleanup_functions`, `for function in functions:`) — missing the `is_deleted_or_deleting()` skip that `extras.py::_discover_function_extras` (line ~116) applies. The two code paths for "what's an unapproved extra" vs. "what to actually delete" have drifted: a function already mid-deletion can be re-targeted by `fn function delete` here (extra API call / possible error), and since one path filters deleting functions first and the other doesn't, they can also disagree about which functions count as "extra" vs. "known." Recommend extracting one shared discovery/classification helper both paths call.

**`oci_cleanup/domains/services.py:191`** (`cleanup_buckets`) — no tenancy-wide reconciliation against `context.managed_resources` for marker-proven buckets outside `context.compartment_id`, unlike the equivalent handling for event rules/streams (`cleanup_connectors_events_streams`) and functions (`cleanup_functions`), which both explicitly re-check `context.managed_resources` for marker-proven resources the compartment-scoped list missed. If a `DatadogManaged`-tagged backfill bucket lives outside the target compartment — the same tenancy-wide-marker scenario the event-rule code comment explicitly calls out — it will silently never be discovered, with no failure surfaced.

**`oci_cleanup/domains/services.py:43`** (and similarly lines 77, 111 for event rules/streams) — `is_owned(x) or defined_marker(x)` has no exact-name check, unlike `exact_owned` used for VCN/subnet/function-app elsewhere in the stack. Freeform tags are customer-editable; a resource that inherited or was manually tagged `ownedby=datadog` would be deleted with no secondary sanity check. Worth aligning with the `exact_owned` bar the rest of the codebase sets for itself.

`_discover_function_extras` (extras.py) and `cleanup_functions` (services.py) duplicate ~45 lines of near-identical application/function discovery logic (list applications → filter `exact_owned` → list functions → compute `marker_proven`). Because they're separately maintained, they've already drifted (see the `is_deleted_or_deleting` gap above). Recommend a single shared helper returning `(application, functions)` pairs, consumed by both the "what needs approval" and "what to delete" code paths.

### Style / simplification

The inline prefix tuple in `cleanup_connectors_events_streams` (`("ocid1.bucket.", "ocid1.eventrule.", "ocid1.fnfunc.", "ocid1.serviceconnector.", "ocid1.stream.")`) has no comment explaining why bucket/fnfunc/serviceconnector are listed as "supported but handled elsewhere." Worth a short comment or hoisting to a module constant, since more domains are likely to be added later in the stack and this list will need to stay in sync.

EamonBrady1 (approved):

griffin-michalak (approved):

PR #168, opened 2026-08-14
Schedule KMS resources for deletion

johannzhang8:
## Summary
- schedule deletion of the Datadog API-key secret before its key and vault
- persist OCI-compliant deletion timestamps so retries remain stable
- report throttling and vault quota remediation clearly

## Stack
- Depends on https://github.com/DataDog/oracle-cloud-integration/pull/167

## Test plan
- [x] `python3 -m unittest discover -s oci-integration-cleanup/tests -p 'test_*.py' -v` (32 tests)

### DIFF
--- a/oci-integration-cleanup/integration_cleanup.py
+++ b/oci-integration-cleanup/integration_cleanup.py
@@ -26,6 +26,7 @@ from oci_cleanup import __all__ as package_api
 from oci_cleanup.base import CleanupBase
 from oci_cleanup.discovery import DiscoveryMixin
 from oci_cleanup.domains.extras import ExtrasMixin
+from oci_cleanup.domains.kms import KmsMixin
 from oci_cleanup.domains.network import NetworkMixin
 from oci_cleanup.domains.services import ServicesMixin
 from oci_cleanup.engine import EngineMixin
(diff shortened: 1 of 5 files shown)


johannzhang8 (commented):
## Review notes

(Posting as a single review comment, not inline — see note on earlier PRs in this stack: this org's EMU/SSO policy blocks diff-anchored comment endpoints for my token.)

This is the cleanest PR in the stack — consistent `exact_owned` usage, correct secret→key→vault deletion ordering, and `_deletion_time` (kms.py:35) correctly persists the computed timestamp into the manifest on first call and reuses it on retries, so retries don't perpetually push the deletion date forward. The shared `kms_deletion_time` computed once for both key and vault (kms.py:51-52) is a nice way to satisfy OCI's "vault time ≥ key time" constraint without extra bookkeeping.

### Style / simplification

**`oci_cleanup/domains/kms.py:115-119`** — `vault.get("management-endpoint") or vault.get("management_endpoint") or ""` hedges hyphen/underscore field naming, where the rest of the codebase has dedicated helpers (`resource_name`/`resource_id`) for this kind of normalization. Minor inconsistency, not a bug — could reuse or extend those helpers here too.

No correctness issues found beyond the general note (shared with earlier PRs in this stack) that `exact_owned` should be the uniform ownership-check bar everywhere destructive actions are taken — this PR already meets that bar consistently.

cale-at-ddog (approved):

PR #169, opened 2026-08-14
Complete stack, identity, and compartment teardown

johannzhang8:
## Summary
- destroy regional Resource Manager stacks before direct orphan cleanup and remove the parent stack last
- remove validated IAM users, groups, policies, dynamic groups, and managed tags only after regional cleanup succeeds
- add guarded compartment teardown as the final cleanup stage: `--delete-compartment` is required, and execution requires `--dry-run false`, a proven Quickstart-created and owned compartment, no residual resources or child compartments, and no pending KMS deletions
- document the complete stateless recovery workflow and safe retry behavior

## Stack
- Depends on https://github.com/DataDog/oracle-cloud-integration/pull/168
- For context on how this package works: [README](https://github.com/DataDog/oracle-cloud-integration/blob/d44f6dd1e62fe0f75c628c2f738e54024d1ede78/oci-integration-cleanup/README.md)

## Test plan
- [x] `python3 -m unittest discover -s oci-integration-cleanup/tests -p 'test_*.py' -v` (64 tests)

### DIFF
--- a/oci-integration-cleanup/oci_cleanup/domains/stacks.py
+++ b/oci-integration-cleanup/oci_cleanup/domains/stacks.py
@@ -4,21 +4,23 @@ Safety boundary: preserves failed stacks for retry and only accepts constrained 
 Cleanup sequence role: handles regional stacks before orphan services and the parent near the end.
 
 ``StacksMixin`` validates region-prefixed child stacks and the selected parent stack,
-starts Resource Manager destroy jobs, and checks terminal job status. A stack is
-recorded complete only after a successful destroy, leaving failures resumable.
+reconciles Resource Manager destroy jobs, and deletes a stack record only after
+OCI reports a successful destroy.
 """
 
 from __future__ import annotations
 
+import datetime as dt
 from typing import Any
 
 from ..constants import LOGGER, REGIONAL_STACK_PREFIX
-from ..errors import CleanupError, raw_error_message
+from ..errors import CleanupError
 from ..models import CleanupContext
 from ..resources import (
     exact_owned,
     is_owned,
     lifecycle_state,
+    resource_field,
     resource_id,
     resource_name,
 )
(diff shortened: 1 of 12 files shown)


johannzhang8 (commented):
## Review notes

(Posting as a single review comment, not inline — see note on earlier PRs in this stack: this org's EMU/SSO policy blocks diff-anchored comment endpoints for my token.)

### Correctness

**`oci_cleanup/domains/stacks.py:134-141`** (`cleanup_parent_stack`) — **no ownership validation before destroying `--parent-stack-id`**:
```python
def cleanup_parent_stack(self, context: CleanupContext) -> None:
    if not self.args.parent_stack_id:
        return
    stack = {"id": self.args.parent_stack_id, "display-name": "explicit parent stack"}
    self.destroy_regional_stack(context.home_region, stack)
```
Every other destructive path in this codebase (identity, tags, compartment, KMS in #168) checks `exact_owned`/`is_owned` before deleting. This one destroys whatever stack ID it's given with zero tag/name check — and the module docstring ("only accepts constrained stack identities") isn't actually true for this path. A fat-fingered `--parent-stack-id` would destroy an arbitrary Resource Manager stack, Datadog-owned or not. This is the one issue in the whole 5-PR stack I'd block on — recommend adding the same ownership check used elsewhere before calling `destroy_regional_stack` here.

**`oci_cleanup/domains/stacks.py:130`** (`cleanup_regional_stacks`) — matches stacks by name-prefix only (`resource_name(stack).startswith(REGIONAL_STACK_PREFIX)`), with no tag/ownership check, unlike the identity/tag cleanup code elsewhere in this same PR. A customer stack that happens to share the naming prefix would be destroyed.

### Style / simplification

**`oci_cleanup/domains/stacks.py:27`** — `cleanup_parent_stack` reuses `destroy_regional_stack`, which always builds the manifest action id as `regional-stack:{region}:{stack_id}` — this makes the parent-stack deletion indistinguishable from a real regional-stack deletion when someone audits the manifest later. Consider a distinct action-id prefix for the parent-stack path.

### Cross-cutting note (applies across the stack)

Ownership-check strength is inconsistent across resource types in this 5-PR stack: PR #166's non-default route tables (`is_owned` only), PR #167's connectors/event-rules/streams/buckets (`is_owned(x) or defined_marker(x)`, no name check), and this PR's regional/parent stacks (name-prefix or no check at all) are all weaker than the `exact_owned` (tag + exact name + compartment) bar used for VCNs, subnets, function apps, and KMS resources elsewhere. Given `exact_owned` is already the established pattern for the highest-risk resources, I'd suggest making it the uniform bar everywhere a destructive action is taken, rather than three different weaker variants scattered across the stack.

oci-integration-cleanup/oci_cleanup/domains/stacks.py
```diff
@@ -0,0 +1,290 @@
...
+            ],
+        )
+        for stack in stacks:
+            name = resource_name(stack)
+            if name.startswith(REGIONAL_STACK_PREFIX) and exact_owned(
+                stack,
+                expected_names={name},
+                compartment_id=context.compartment_id,
+            ):
+                self.destroy_regional_stack(region, stack)
```
rheei: This ownership check can never pass for any regional stack, because the freeform tag it depends on is never set anywhere in the merged code. Commit ad6bc6ba added `--freeform-tags '${jsonencode(local.tags)}'` (`local.tags = {ownedby = "datadog"}`) to the `oci resource-manager stack create`/`stack update` calls in [`datadog-integration/regional_stack.tf`](https://github.com/DataDog/oracle-cloud-integration/blob/d38a581b4f2cecc87b943e7f4d8c3d4bbc958811/datadog-integration/regional_stack.tf#L58-L79) as the tagging mechanism for this exact `exact_owned()` check, but two commits later `86d00ede` reverted those two `--freeform-tags` lines with no replacement — and that revert is still in effect at HEAD (neither of the two commits since has touched `regional_stack.tf` or `locals.tf`). `is_owned()`/`exact_owned()` (`oci_cleanup/resources.py`) read freeform tags only, with no fallback. So every regional stack created/updated going forward will lack `ownedby=datadog`, and `cleanup_regional_stacks` will always skip destroying it (same issue applies to `cleanup_parent_stack`'s `is_owned(stack)` check below, lines 278-284) — silently defeating this PR's own stated goal.

🤖 Generated with [Claude Code](https://claude.ai/code)

rheei (approved):

PR #171, opened 2026-08-17
Discover and approve unexpected cleanup resources

johannzhang8:
## Summary
- discover unexpected nested network and compute resources before regional cleanup
- require explicit per-run approval and fail closed without an interactive terminal
- add OCI field normalization and focused approval, VNIC, and compartment discovery coverage

## Test plan
- [x] `python3 -m unittest discover -s oci-integration-cleanup/tests -p 'test_*.py' -v`
- [x] `python3 oci-integration-cleanup/integration_cleanup.py --help`
- [x] `git diff --check`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/oci-integration-cleanup/oci_cleanup/domains/extras.py
+++ b/oci-integration-cleanup/oci_cleanup/domains/extras.py
@@ -1,0 +1,747 @@
+"""Responsibility: discover, confirm, and delete unexpected nested resources.
+
+Safety boundary: fails closed for non-interactive input and persists explicit approvals.
+Cleanup sequence role: runs after discovery and before concurrent regional cleanup.
+
+``ExtrasMixin`` finds unexpected functions, VNIC attachments, and network dependents within 
+the Datadog created function apps and VCN resources and
+turns each into an auditable candidate command, and prompts on every execute run
+while the blocker remains live. Later stages use the current run's decisions to
+delete approved children or preserve their parents.
+"""
+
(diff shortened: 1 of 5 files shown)


oci-integration-cleanup/oci_cleanup/domains/extras.py
```diff
@@ -0,0 +1,747 @@
...
+                candidate.container_name,
+                candidate.kind,
+                candidate.resource_id,
+            ),
+        )
+
+    def _ask_yes_no(self, prompt: str) -> bool:
+        stream = sys.stdin
+        if not stream.isatty():
+            LOGGER.warning("%s [y/n]: no interactive TTY; defaulting to n", prompt)
```
EamonBrady1: does this mean we give customers the option to manually approve every deletion? I definitely think that's something this script needs to have, otherwise no customer will trust using it

EamonBrady1 (commented):
Are we cleaning up connector hubs and events functions that are provisioned by hubmanager? What about backfill buckets?

oci-integration-cleanup/oci_cleanup/domains/extras.py
johannzhang8: yes, any resource that the customer makes within our resources will push a prompt of whether or not they will want to delete it

EamonBrady1 (approved):

